PR #4, opened 2021-03-10
Added appTheme.ts to support design system "bases"

hthomas-czi:
* Includes colors, textColors, spacing, corners, and primaryAppFont
* Still needs typography, borders, and elevation


src/core/styles/app/appTheme.ts
```diff
@@ -0,0 +1,76 @@
+export const colors = {
```
ovalenzuela19, line 1: One concern Harley and I talked about yesterday was how to incorporate the the `appTheme` with the `defaultTheme`. 

- What's the difference between them?
- When should we use `appTheme`, `defaultTheme`, or both? 
- How should we use `appTheme`?
- Would other teams have to define their own `appTheme` & `defaultTheme`?

tihuan: Great questions!

Sounds like we need to figure out which one should be the source of truth?

From the trend I'm seeing, it feels like the world is moving towards using the computed theme (i.e. `defaultTheme` instead of `appTheme`) as the source of truth, and incorporate it with a design system library, such as `@material-ui/system` to style each component accordingly

Doc: https://material-ui.com/system/basics/

Intro: https://medium.com/material-ui/introducing-material-ui-design-system-93e921beb8df

How system works with spacing: https://material-ui.com/system/spacing/

HOWEVER, at Meta, we started out using `defaultTheme` as the source of truth, and then eventually moved back to using `appTheme` as the source of truth, because `appTheme` comes with better type support and we didn't have `@material-ui/system` back then, which tremendously helps with styling using t`defaultTheme`

So I think I'm definitely open to trying using @material-ui/system and `defaultTheme` to style components!

ovalenzuela19: Sounds good to me! What if there's some styling that the `defaultTheme` does not have, should we include styling from `appTheme` ad-hoc?

seve:
I'm going to disable the order keys lint rule as it's slowing us down a little @tihuan

src/core/styles/app/appTheme.ts
tihuan: Wow I've been playing with `@material-ui/system` + Emotion. It's SO CONVENIENT!!!

Essentially, `@material-ui/system` uses Decorator pattern to enhance a styled component to expose extra API surface that taps into `defaultTheme` to set styling values

In the following example, I added `palette` and `shadows` from `@material-ui/system` to `StyledAlert`, and now we get props `boxShadow`, `color`, and `bgcolor` that we get to play with.

Setting `bgcolor={info.main}` means using the color from `defaultTheme.palette.info.main`
and `boxShadow={15}` means using `defaultTheme.shadows[15]`

Really cool!!

https://codesandbox.io/s/material-demo-forked-jkb49?file=/demo.tsx

Screen Shot 2021-03-10 at 11 43 40 AM

hthomas-czi: Teams would define their own `appTheme` which would allow teams to define their own styles (colors, spacing, etc.). `defaultTheme` would be exclusively constructed from `appTheme` so that any app-specific style overrides are automatically exposed as Material-UI expects them (e.g., `palette.primary.main`). 

The purpose of having a separate `appTheme` is that — as far as I can tell — the `ThemeOptions` object in `defaultTheme` doesn't support the level of granularity that our design system has. For example, there are only 4 variants per color (`dark`, `light`, `main`, and `contrastText`) while the IDSeq design system has 6 `primary` colors — as would the cellxgene design system. 

We could consider just ignoring `defaultTheme` and use `appTheme` directly, but I was trying to avoid breaking everything because I believe Material-UI's core components expect those values to be defined. However, this introduces a strange inconsistency in how we pass styles to components. Do you use `appTheme` or `palette`? Thoughts @tihuan @ovalenzuela19?

tihuan:
Oh for object sorting I use https://marketplace.visualstudio.com/items?itemName=zengxingxin.sort-js-object-keys!

Let me pull the branch to fix the ordering real quick

seve:
@tihuan OH MY GOSH I'VE BEEN DOING IT MANUALLY FOR MONTHS

seve:
I wonder if we can append to the error message with that extension

jshoe:
Gotta keep your ABCs sharp

tihuan:
https://github.com/chanzuckerberg/sci-components/pull/6

Ahh sorry should have shared the extension with you much sooner, @seve 😆 I also only discovered it after manually sorting for months too 🤦

tihuan (approved):
LOOKS SUPER NEAT!! Thanks so much for adding this, @hthomas-czi ⭐ 🔥 🚀 !!

PR #5, opened 2021-03-10
Add examples for Button and Checkbox

jshoe:


package.json
```diff
@@ -67,5 +69,6 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx,md,html,css,yml}": "npx prettier --write"
-  }
+  },
+  "dependencies": {}
```
ovalenzuela19, line 73: was an empty `dependencies` generated automatically?

ovalenzuela19 (approved):

package.json
jshoe: No I accidentally added and then removed a package

PR #7, opened 2021-03-10
Thuang dropdown item multi select columns

tihuan:
Screen Shot 2021-03-11 at 4 43 53 PM

demo

### DIFF
--- a/src/core/styles/common/constants/black.ts
+++ b/src/core/styles/common/constants/black.ts
@@ -1,0 +1,3 @@
+export enum Black {
+  DEFAULT = "#000000",
+}
--- a/src/core/styles/common/mixins/body.ts
+++ b/src/core/styles/common/mixins/body.ts
@@ -1,0 +1,9 @@
+import { css } from "@emotion/react";
+import { FontWeights } from "../constants/fontWeights";
+
+export const fontXs = css`
+  font-size: 13px;
+  line-height: 20px;
+  letter-spacing: 0.3px;
+  font-weight: ${FontWeights.REGULAR};
+`;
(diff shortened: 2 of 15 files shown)


src/core/styles/common/mixins/body.ts
```diff
@@ -0,0 +1,9 @@
+import { css } from "@emotion/react";
+import { FontWeights } from "../constants/fontWeights";
+
+export const fontXs = css`
```
tihuan, line 4: @hthomas-czi this is an example of defining a mixin!

tihuan: And this is how to access the theme object in Emotion 😄 

```ts
const StyledAlert = styled(Alert)`
  ${shadows}
  ${palette}
  ${({ theme }) => {
    console.log("!!!! GOT THEME", theme);
    return `
      background-color: ${theme.palette.success.light}
    `;
  }}
`;
```

https://codesandbox.io/s/material-demo-forked-jkb49?file=/demo.tsx:782-988

Screen Shot 2021-03-10 at 2 00 39 PM

src/core/styles/common/constants/black.ts
```diff
@@ -0,0 +1,3 @@
+export enum Black {
```
ovalenzuela19, line 1: General question: Do we need these constants defined? Aren't these supposed to be defined in the `appTheme`?

tihuan: Ahh that's right 😆 ! Totally forgot about that lol Thanks for catching this!

Yeah we will need to pass `theme` to `Emotion` and then use it for styling!

We can do it after Create-a-thon when we have time to review everything!

jshoe (approved):
Wow, nice!

PR #11, opened 2021-03-11
Add flat and elevated alerts

j-x-han:
**Flat alerts:**
image

**Elevated alerts**
image

**Snackbox alert test**
ezgif com-gif-maker (7)

image


src/core/Alert/index.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+          vertical: "top",
+          horizontal: "right",
+        }}
+        open={open}
+        autoHideDuration={6000}
+      >
+        <StyledAlert className="elevated" severity="info">
+          <div>This is a snackbar alert!</div>
+          <DismissButton onClick={handleClose}>DISMISS</DismissButton>
+        </StyledAlert>
```
j-x-han, line 65: @tihuan I went with this instead of the `makeStyles` approach you suggested to position the Dismiss button, since making the root alert container's `flex-direction: "column"` messes with the icon positioning 😅 
image
versus
image

src/core/Alert/index.stories.tsx
```diff
@@ -0,0 +1,124 @@
+import styled from "@emotion/styled";
+import { Snackbar } from "@material-ui/core";
+import CheckCircleOutlineIcon from "@material-ui/icons/CheckCircleOutline";
+import ErrorOutlineIcon from "@material-ui/icons/ErrorOutline";
+import { storiesOf } from "@storybook/react";
+import React from "react";
+import Button from "../Button";
```
j-x-han, line 7: @tihuan Am I importing our Button incorrectly? I'm getting a lot of these errors (but not when I use the base material-ui button)

image

ovalenzuela19:
THIS LOOKS AMAZING

jshoe (approved):

package-lock.json
```diff
@@ -0,0 +1,16962 @@
+{
```
jshoe, line 1: You don't need to add this since there's yarn-lock.json actually

j-x-han: Got it, removed the file!

PR #12, opened 2021-03-11
Added typography, shadows, and beta colors to appTheme

hthomas-czi:

PR #13, opened 2021-03-11
defaultTheme-add-shadows

tihuan:


jshoe (approved):

tihuan:
Thanks @jshoe !!

hthomas-czi (commented):
Hey @tihuan!

I actually added shadows (plus some other stuff) in [this PR](https://github.com/chanzuckerberg/sci-components/pull/12). What do you think about using that instead of this?

PR #15, opened 2021-03-12
feat(component): Create Tabs

seve:
This isn't feature complete but definitely demoable.

A Major change is removing the automatic action creation based on regex match.  This was removing any controlled functionality connected to event handling.

Adds @emotion/css import to allow exporting a css definition as a class name

image
image


tihuan (approved):
Looks fantastic!! LGTMMM 🥇 🔥 ✨ !

PR #19, opened 2021-04-29
Add List and Link

tihuan:
1. Add and export List
2. Add and export Link
3. Export more mixins 

PTAL thank you 😄 🎉 

Screen Shot 2021-04-29 at 7 57 49 AM

Screen Shot 2021-04-29 at 7 58 08 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "czifui",
-  "version": "0.0.18",
+  "version": "0.0.25",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "repository": {
--- a/src/core/Link/index.stories.tsx
+++ b/src/core/Link/index.stories.tsx
@@ -1,0 +1,22 @@
+import { Args, Story } from "@storybook/react";
+import React from "react";
+import Link from "./index";
+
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "czifui",
-  "version": "0.0.18",
+  "version": "0.0.25",
```
jshoe, line 3: I'm just curious, why did it go from 18 to 25?

jshoe (approved):
Looks good to me!

package.json
tihuan: Great question! It's a long story lol

So Aspen uses happy path dev environment for local development now, which mounts the project directory in Docker local VM. BUT that doesn't mount the component library directory, so local project can't use the local component library

Thus, the only way for local project to use the component library is to publish it to `npm` once in a while 😆 

I try to decrease the publish frequency by testing things more thoroughly in Storybook. But sometimes I could brain fart and publish a new version, forgetting to export the new component first, thus needing to export, rebuild the package, bump the version, and publish again lol

I'm sure we can figure out a better process 😆

tihuan:
Thanks again for reviewing, @jshoe ! Merging!

PR #22, opened 2021-05-06
thuang-tooltip

tihuan:


src/core/Tooltip/index.tsx
```diff
@@ -1,39 +1,16 @@
...
-} from "@material-ui/core";
+import { useTheme } from "@emotion/react";
+import { Tooltip as RawTooltip, TooltipProps } from "@material-ui/core";
 import React from "react";
-import { AppThemeOptions } from "../styles";
+import { arrowCss, tooltipCss } from "./style";
 
 export { TooltipProps };
 
-const useStyles = makeStyles((theme: AppThemeOptions) => {
```
tihuan, line 11: `makeStyles()` generates a css class name for each key (e.g., `arrow` and `tooltip`), so we can replace it with [`css()`](https://emotion.sh/docs/@emotion/css) from `import { css } from '@emotion/css`, which also generates a css class name as an output!

Note that there's ANOTHER `css()`, which is from `import { css } from '@emotion/react'`, which the output is serialized style instead of css class name. So we just need to be aware of when to use which, this confused me so much at the beginning 😂 

`themeToCss()` in `src/core/styles/common/mixins/fonts.ts` is an example of using `css()` from `@emotion/react`!

src/core/Tooltip/style.ts
```diff
@@ -0,0 +1,37 @@
...
+  AppThemeOptions,
+  fontBodyXs,
+  getColors,
+  getShadows,
+  getSpacings,
+} from "../styles";
+
+export const tooltipCss = (theme: AppThemeOptions): string => {
+  const props = { theme };
+  const colors = getColors(props);
```
tihuan, line 12: since all the selectors expects a `props` interface, which is an object with key `theme`, we need to declare a `props` variable and pass it to all the selectors

src/core/Tooltip/style.ts
```diff
@@ -0,0 +1,37 @@
...
+  const shadows = getShadows(props);
+  const spacings = getSpacings(props);
+
+  return css`
+    ${fontBodyXs(props)}
```
tihuan, line 17: Here we also need to call `fontBodyXs()` with `props`, because `css()` from `@emotion/css` doesn't call `fontBodyXs()` with props like `@emotion/styled` does

tihuan (commented):
I think using `@emotion/styled` doesn't work for Tooltip component, because MUI Tooltip doesn't forward the css class name `styled` generates to its rendered elements, since the Tooltip's [css interface](https://material-ui.com/api/tooltip/#css) doesn't include `root`, so Tooltip wouldn't know where to forward the `styled` class name 😄 !

On the other hand, Button's [css interface](https://material-ui.com/api/button/#css) has `root`, so `styled(Button)` should work!

PTAL thank you 🤩 🙏 !

j-x-han (approved):
This looks awesome! Thanks Timmy!

src/core/Tooltip/style.ts
j-x-han: Oh interesting! If I also wanted the `Tooltip` component to take in some props that would affect styling, what would be the best way to go about doing that?

For context, Jenn had just reached out to me that they had updated some of the naming for the Tooltips which you can find [here](https://www.figma.com/file/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=642%3A4116), but essentially `TooltipInfo` is now being consolidated into just `Tooltip` as its secondary style. `Tooltip`'s primary style will be inverted (black background, white text).

tihuan: Got it! What do you think of new prop `inverted` for that? 

From API's perspective, I feel "primary" and "secondary" are closely related to color. Like MUI's `color` prop typically accepts string `"primary"` and `"secondary"` to style the component with primary color or secondary color. So since what we're doing is not related to primary and secondary colors, I feel steering away from those terms would avoid confusion 😆  E.g., [Chip](https://material-ui.com/api/chip/#props)

As for accepting custom props, typically I use `interface ExtraProps { ... }` to capture them, and then update the component prop interface like so:

```ts
// src/core/ListItem/index.tsx

import { ListItemProps as RawListItemProps } from "@material-ui/core";
import { ExtraProps } from "./style";

export type ListItemProps = RawListItemProps & ExtraProps;

const ListItem = (props: ListItemProps): JSX.Element => {
  return <StyledListItem disableGutters {...(props as never)} />;
};
```

In our Tooltip case, it could be something like the following? :

```tsx
interface ExtraProps {
  inverted?: boolean;
}

type TooltipProps = ExtraProps & RawTooltipProps;

const TooltipInfo = (props: TooltipProps): JSX.Element => {
  const theme = useTheme();
  const tooltip = tooltipCss({ ...props, theme });
  const arrow = arrowCss({ ...props, theme });

  return <RawTooltip arrow classes={{ arrow, tooltip }} {...props} />;
};

TooltipInfo.defaultProps = {
  inverted: false,
};
```

j-x-han: This looks great! Yeah, I'm also more accustomed to using `inverted` from semantic-ui as well. I'll ask Jenn what she thinks about changing it to inverted!

PR #23, opened 2021-05-06
thuang-tooltip-2

tihuan:


src/core/Tooltip/style.ts
```diff
@@ -1,17 +1,19 @@
...
-export const tooltipCss = (
-  { inverted = false },
-  theme: AppThemeOptions
-): string => {
-  const props = { theme };
+export interface ExtraProps extends Props {
+  inverted?: boolean;
+}
+
+export const tooltipCss = (props: ExtraProps): string => {
```
tihuan, line 14: I updated the css class helper functions to consolidate the props to accept just one argument `extraProps`, so the interface is identical to how we use `styled` to write CSS!

j-x-han (approved):
Awesome! Thanks for these fixes 😄

PR #25, opened 2021-05-10
add Chips and export fontCaps mixins

tihuan:
1. Add chip (or tag as called in our design system)
2. Export `fontCaps` mixins
3. Add Tooltip styling arrow example 

Screen Shot 2021-05-10 at 10 08 46 AM
Screen Shot 2021-05-10 at 10 08 50 AM
Screen Shot 2021-05-10 at 10 08 53 AM
Screen Shot 2021-05-10 at 10 08 55 AM
Screen Shot 2021-05-10 at 12 50 21 PM
Screen Shot 2021-05-10 at 12 50 24 PM

demo

### DIFF
--- a/src/core/Tooltip/index.stories.tsx
+++ b/src/core/Tooltip/index.stories.tsx
@@ -1,3 +1,4 @@
+import { css } from "@emotion/css";
 import InfoOutlinedIcon from "@material-ui/icons/InfoOutlined";
 import { Args, Story } from "@storybook/react";
 import React from "react";
@@ -31,4 +32,14 @@ export const Info = Template.bind({});
 
 Info.args = {
   inverted: false,
+};
+
+export const StyledArrow = Template.bind({});
+
+const arrow = css`
+  left: 0 !important;
+`;
+
+StyledArrow.args = {
+  classes: { arrow },
 };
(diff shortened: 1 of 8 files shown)


src/core/Tooltip/index.stories.tsx
```diff
@@ -32,3 +33,13 @@ export const Info = Template.bind({});
 Info.args = {
   inverted: false,
 };
+
+export const StyledArrow = Template.bind({});
+
+const arrow = css`
```
tihuan, line 39: this demonstrates that consumer can add extra styles to arrow as needed

src/core/styles/common/mixins/fonts.ts
```diff
@@ -26,6 +26,29 @@ export const fontBodyXs = fontBody("xs");
...
+    } = typography;
+
+    return css`
+      ${themeToCss(caps[fontSize])}
+      text-transform: uppercase;
+    `;
+  };
+};
+
+export const fontCapsXxs = fontCaps("xxs");
```
tihuan, line 48: Chip uses this mixin, so might as well export them

tihuan (commented):
PTAL thank you!!

jshoe (approved):
Haha I like the name chips!

src/core/styles/common/mixins/fonts.ts
j-x-han: Nice! I end up using this for `TooltipTable` as well, so thanks for adding! 😄

tihuan: Amazing!! Go team 🚀 !!

PR #27, opened 2021-05-11
thuang-tooltip

tihuan:


src/core/TooltipTableContent/index.tsx
```diff
@@ -1,19 +1,15 @@
...
 import {
-  alertCss,
-  disabledCss,
+  Alert,
   ExtraProps,
-  rowLabelCss,
-  rowValueCss,
-  sectionCss,
-  sectionLabelCss,
+  RowLabel,
```
tihuan, line 9: I think MUI's recommendation is to use `@emotion/styled` to style components by default, and only use `@emotion/css` to generate css class names when we need to pass class names to an MUI's component's `classes` props!

https://emotion.sh/docs/styled

This also means we no longer need `classnames` package 🎉

src/core/TooltipTableContent/index.tsx
```diff
@@ -23,52 +19,27 @@ export { TooltipTableContentProps };
...
 
-  const theme = useTheme();
-
-  const extraProps = {
-    /* stylelint-disable property-no-unknown -- false positive */
-    theme,
-    /* stylelint-enable property-no-unknown -- false positive */
-  };
-
-  const alertClassName = alertCss(extraProps);
```
tihuan, line 34: instead of generating these CSS classnames, we can generate styled components directly!

E.g.,

`alertCss(extraProps)` -> `Alert` component

```ts
export const Alert = styled.div`
  ${fontBodyXxs}
`;
```

And `rowLabelCss` -> `<RowLabel />`

```ts
export const RowLabel = styled(TableCell)`
  ${fontHeaderXs}
  ${disabledCss}

  padding: 0;
  width: 50%;
`;
```

tihuan (commented):
Mainly just refactored using `@emotion/styled` 😄 PTAL! Thank you 🚀 !!

j-x-han (approved):
Super informative! Thanks as always Timmy 😄

tihuan:
Always a pleasure to help, @j-x-han 🎉 !! GO TEAM!!

PR #39, opened 2021-07-30
[CH-148810] Implement large square Tag

j-x-han:
## Description

Adds ~~Tag~~ Chip (size: large, variant: square)

Consolidated into the existing Chip component. Storybook demos actually say "Chip" instead of "Tag" now, but otherwise the following screenshots are still accurate.

Screen Shot 2021-07-29 at 5 21 26 PM

ezgif com-gif-maker (13)

**Design Link:** https://www.figma.com/file/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=642%3A3902

## Notes
* Added props `size` and `variant` but the other size/variants aren't implemented yet.
* Currently just using the material-ui icon for the close icon on the tag. Should be replaced by the SDS icon once we decide if/how we want to import all our custom icons.
* Prior to merging, will run `yarn build && npm version patch && npm publish`.


tihuan:
Hi @j-x-han ! Sorry only got time to check your PR today 😆 

We have Chip component in the library too, maybe there's a way to consolidate both components?

Thanks so much!

j-x-han:
> Hi @j-x-han ! Sorry only got time to check your PR today 😆
> 
> We have Chip component in the library too, maybe there's a way to consolidate both components?
> 
> Thanks so much!

No worries! Yeah, I'll look into consolidating them; I'm most likely going to merge the Chip styles into Tag so that we can have the component names match the design system! Hopefully this won't be too problematic if products are already using the Chip component.

tihuan:
Yeah that makes me wonder if we should change the Design System to use Chip instead of Tag, so it's more consistent with MUI's naming 🤔  Let me ask people in Slack!

j-x-han:
I've consolidated my Tag changes into Chip for now! Subject to change depending on the response in Slack 😄 
Screen Shot 2021-08-04 at 3 40 40 PM

tihuan:
Amazing! Thanks so much for updating so quickly!

One last thing is the API `dismissable`, it looks like MUI uses `onDelete`, which shows the delete icon when `onDelete()` is defined. Do you think we should follow that?

https://material-ui.com/api/chip/#props

Thanks again!

j-x-han:
Just updated to remove `dismissable` and use `onDelete` instead!

I also just noticed these console errors, do you have advice on how to avoid these?

Screen Shot 2021-08-04 at 4 09 17 PM

tihuan:
Yay go @j-x-han !! Thanks so much 🙏 
Let me fix the typing stuff real quick!

tihuan:
HMM actually, it seems like SDS only has large and small, which is similar to MUI's medium and small, so we could probably just change `large` to `medium`!

And for rounded/square, MUI doesn't have an API for that, so we'll need to come up with our own. 

`variant` seems to be used for the component general style, like `outlined`, `contained`, `text`, so `square` doesn't fit that criteria, sorry for missing that in review 😆 

I'm thinking about using `isRounded`  here, and default to `true`, since MUI's Chip is rounded by default. So square tag will be `isRounded={false}`

What do you think?

j-x-han:
Those changes sound good to me! I'll push a commit for them, and also follow up in the Slack thread with the designers to see how they feel about these naming changes as well!

j-x-han:
One last error, `isRounded` is somehow getting passed down to the DOM element:
Screen Shot 2021-08-04 at 5 09 08 PM

tihuan:
Whoohoo so close!

Yes `src/core/List/style.ts:11` has a pattern to not forward that prop 🙆‍♂️ 

```tsx
// (thuang): Please keep this in sync with the props used in `ExtraProps`
const doNotForwardProps = ["ordered"];

export const StyledList = styled(List, {
  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
})`
  ${(props: ExtraProps) => {
    if (!props.ordered) return "";

    return `
      counter-reset: section;
    `;
  }}
`;
```

j-x-han:
Ah perfect thank you so much!! Just fixed 🎉

tihuan:
Yayyy amazing! LGTM 🔥 🔥 🔥

tihuan (approved):
LGTM!! 🎉🎉🎉

Janeece (commented):
From my perspective this all looks correct to me. Definitely will be good to get harley's review to make sure its synced with the props in his ref page.

Janeece:
@j-x-han Just to confirm, the border on the tag is just an artifact of the gif, right? I didnt see any borders in the files, but its good to check :)

j-x-han:
@Janeece Yeah I believe the border is either a gif artifact or just an older screenshot I took. The screenshot in the comments that says "Chip" instead of "Tag" is the most updated screenshot, and there isn't a border there.

PR #40, opened 2021-07-31
Thuang add ref to button

tihuan:
This PR is needed to work with our Tooltip component per https://material-ui.com/components/tooltips/#custom-child-element. Essentially if we supply a child element to Tooltip that's NOT MUI component, we'll need to forward ref 💡 😄 👌 !

PTAL thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "czifui",
-  "version": "0.0.41",
+  "version": "0.0.42",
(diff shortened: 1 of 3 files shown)


phoenixAja (approved):
Looks good! Thanks for the help!!

tihuan:
Anytime, @phoenixAja ! Thanks for spotting this use case!

PR #45, opened 2021-08-20
Fix svg type declaration.

ehoops:
In svgr, the react component is not the default export.  Assuming the default incorrectly was causing storybook not to load.  Now the type definition is accurate and imports need to use `import {ReactComponent as SomeSvgName}...`


tihuan (approved):
Ahh didn't know that! Thanks for the quick fix, @ehoops 🙏 LGTMM!

PR #52, opened 2021-09-02
[ch148881] Implement Radio Button

mdjmoore:
[ticket link](https://app.clubhouse.io/sci-design-system/story/148881/implement-radio-button)

Implement the SDS Radio Button in Storybook. 

See attached for checked and unchecked stages, hover and disabled states
https://user-images.githubusercontent.com/35696944/131895259-c2401657-1f3d-45a8-90af-1c9e125ef83c.mov


src/common/svgs/IconRadioChecked.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8 16C12.4183 16 16 12.4183 16 8C16 3.58172 12.4183 0 8 0C3.58172 0 0 3.58172 0 8C0 12.4183 3.58172 16 8 16Z" fill="white"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M16 8C16 12.4183 12.4183 16 8 16C3.58172 16 0 12.4183 0 8C0 3.58172 3.58172 0 8 0C12.4183 0 16 3.58172 16 8ZM11 8C11 9.65685 9.65685 11 8 11C6.34315 11 5 9.65685 5 8C5 6.34315 6.34315 5 8 5C9.65685 5 11 6.34315 11 8Z" fill="#000"/>
```
mdjmoore, line 3: original file had the fill color for both of these as `#3867FA` which meant that color overrides weren't working as expected. Hard coded these to be black so the `.svgrrc` functions as expected

src/core/Radio/index.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <RadioButton
+      disabled
+      inputProps={{ "aria-label": "checked and disabled" }}
+      onChange={actions.onChange}
+      stage="checked"
+    />
+  </div>
+));
+
+// storiesOf("Checkbox", module).add("indeterminate", () => (
```
tihuan, line 79: is this dead code we can remove 😄 ?

src/core/Radio/index.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+      inputProps={{ "aria-label": "radio button unchecked and disabled" }}
+      onChange={actions.onChange}
+      stage="unchecked"
+    />
+  </div>
+));
+
+storiesOf("Radio Button", module).add("checked", () => (
+  <div style={storyColumn as React.CSSProperties}>
+    <FormControlLabel
```
tihuan, line 54: @mdjmoore can you help investigate a little on what's causing the text and radio button to be misaligned? Thanks so much!!

Screen Shot 2021-09-07 at 11 32 27 AM

tihuan (commented):
Looks really good! PTAL on the misalignment I saw in the video 🙏 Thank you!

src/core/Radio/index.stories.tsx
mdjmoore: my bad!!

src/core/Radio/index.stories.tsx
mdjmoore: found the problem. Mui was changing the viewbox from `0 0 16 16` to `0 0 24 24` when loading the svg, meaning it sat in the upper left of a larger box. Hard coding this in at the component level is the first part of the solution. I also had to define a set height and width for the icon-root, as seen in the `IconButton` css. 

The same problem can be seen on the checkbox:
Screen Shot 2021-09-07 at 3 18 34 PM
I will make another ticket for this. We need to make sure that addressing this bug doesn't push a breaking change to projects that have implemented the checkbox component. 

16 x 16 is not a size that exists in our default theme yet, but is used for both icon inputs. I propose adding it to the iconSizes definition in the default theme:

```  iconSizes: {
    input: { height: 16, width: 16 },
    l: { height: 22, width: 22 },
    s: { height: 14, width: 14 },
    xl: { height: 32, width: 32 },
    xs: { height: 10, width: 10 },
  },

happyimadesignr:
Following! @Janeece too

tihuan:
I see! Thanks for figuring this out 🙏 

Does input icon size being `16px` make sense for design? (I'm assuming design set 16px icon size in the first place, but just wanted to double check 😆 ) CC: @happyimadesignr @Janeece 

Thank you!

Janeece:
@tihuan I think design wants it to be 16px. 

For background: the 16px size is the original size spec for these inputs, before we realized that they are implemented using icons. Design has decided to treat the radio/checkbox icons as separate from other icons and don't follow the same standards (I'd think including sizing).

src/core/Radio/index.stories.tsx
Janeece: I think adding input icon as a size is an interesting idea.

tihuan:
Ooh that totally makes sense now 🙆‍♂️ Thanks so much for the context, @Janeece !

Sounds like adding `input: 16` to the icon size object is a good idea, and perhaps @mdjmoore we just need to add comment there that `input` size is typically reserved for input related icons like checkbox and radio, and not for other icons?

Thank you!

mdjmoore:
@tihuan I've made a branch to add the fix for the checkbox and add the size to the defaultTheme, see https://github.com/chanzuckerberg/sci-components/blob/meaganmoore/hotfix/checkbox-alignment/src/core/styles/common/defaultTheme.ts#L110

If this is okay'd, the plan would be merge that, grab the iconSizes for the radio button and then everything should be okay. My concern is that the checkbox component is in use in a project and this change will affect how they have implemented it. Devs likely would have added their own padding/positioning to correct for the misalignment

tihuan (approved):
LGTM!! Thanks for fixing both Checkbox and Radio, @mdjmoore 🙏 🚀 🔥 You rock!!

PR #53, opened 2021-09-07
Thuang fix alert

tihuan:
This PR fixes Alert component using the component library's default theme instead of the product's default theme 😆 

PTAL thank you!!

### DIFF
--- a/src/core/Alert/style.ts
+++ b/src/core/Alert/style.ts
@@ -1,30 +1,28 @@
 import styled from "@emotion/styled";
 import { Alert } from "@material-ui/lab";
+import { getColors, getShadows, getSpacings } from "../styles";
 import { defaultTheme } from "../styles/common/defaultTheme";
 
 export const StyledAlert = styled(Alert)`
-  margin: ${defaultTheme.spacing(5)}px 0;
-  border-radius: ${defaultTheme.spacing(1)};
-  color: ${defaultTheme.palette.text.primary};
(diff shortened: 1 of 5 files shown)


src/core/Alert/style.ts
```diff
@@ -1,30 +1,28 @@
...
-  margin: ${defaultTheme.spacing(5)}px 0;
-  border-radius: ${defaultTheme.spacing(1)};
-  color: ${defaultTheme.palette.text.primary};
-  padding: ${defaultTheme.spacing(6)}px ${defaultTheme.spacing(6)}px
-    ${defaultTheme.spacing(6)}px 9px;
+  ${(props) => {
+    const colors = getColors(props);
+    const spacings = getSpacings(props);
+    const shadows = getShadows(props);
+    const { severity = "primary" } = props;
```
phoenixAja, line 11: cool i didn't realize that you could unpack an object like this!

phoenixAja (approved):
TY for addressing this so quickly timmy!!

src/core/Alert/style.ts
tihuan: Yaas it works for arrays too!

`const [a, b] = [1, 2]`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

tihuan:
Thanks for the quick review, @phoenixAja !! Merging!

PR #56, opened 2021-09-08
hotfix checkbox alignment

mdjmoore:
[ticket link](https://app.clubhouse.io/sci-design-system/story/160102/hotfix-align-label-for-svg-inputs)

See comment [here](https://github.com/chanzuckerberg/sci-components/pull/52#discussion_r703744174). The misalignment on the radio button is also seen on the checkbox icons. 

This ticket adds an input object to the icon size object in the default theme and adds new css rules to ensure the correct size is used for the icon

**Because the checkbox is already implemented, this change could cause shifts in layout once it's released.**

Screen Shot 2021-09-08 at 10 18 16 AM
Screen Shot 2021-09-08 at 10 18 24 AM
Screen Shot 2021-09-08 at 10 18 29 AM

### DIFF
--- a/src/core/Checkbox/index.tsx
+++ b/src/core/Checkbox/index.tsx
@@ -43,11 +43,25 @@ const Checkbox = (props: CheckboxProps): JSX.Element => {
     <StyledCheckbox
       {...newProps}
       checkedIcon={
-        <SvgIcon fillContrast="white" component={IconCheckboxChecked} />
+        <SvgIcon
+          fillContrast="white"
+          component={IconCheckboxChecked}
+          viewBox="0 0 16 16"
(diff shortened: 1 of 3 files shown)


Janeece:
@mdjmoore This alignment looks good to me. I wonder if we should post about it in the collab channel to see if it'll impact any of the work other engineers have done using checkboxes/give them a heads up it may cause layout shifts.

I'm not really sure if the checkbox has been used in any products... possibly Erin used it for RAO?

Janeece (approved):
Looks good to me!

tihuan (approved):
LGTM too 🙆‍♂️ Thank you!!

PR #57, opened 2021-09-09
[ch155231] callout update

mdjmoore:
**Aspen uses the former `Alert` component, both flat (Callout Basic) and elevated (Notification). This PR should not be merged until the Notification component is added to Storybook and the transition from `Alert` to `Callout` & `Notification` is clear**

[ticket link](https://app.clubhouse.io/sci-design-system/story/155231/update-basic-callout-component-for-sds)

update the Alert component:
- [x] Updated component naming (as in the design documentation).
- [x] Updated text size to fontBodyXs/400.
- [x] Updated container padding to spacingM (10px).
- [x] Updated icon sizing to be iconSizeL.
- [x] Updated all icons to the "Alert" icon (used for all the various intents).
- [x] Updated icon right margin to be spacingM (10px).
- [x] Updated "Close" icon to be iconSizeS, gray500.

Screen Shot 2021-09-15 at 1 18 52 PM
Screen Shot 2021-09-15 at 1 18 59 PM


karenliang0:
Hey Meagan! I am so sorry for the confusion here. We broke up the Alert component into two different components: the "Callout" component (which has "Callout Basic" and "Callout Accordion") and the "Notifications" component. 

For the "Callout" component which is what this ticket is for, it maps to the components that are currently within the "Alert > Default" and "Alert > Flat." 

From the image you attached above, those components are from "Alert > Snackbar" and "Alert > Elevated" which map to the new "Notifications" component instead. 

Here's an image showing the differences between the "Callout" component and the "Notification" component:  
image

Let me know if you want to get in a call Monday and talk through this because it's definitely confusing!!

karenliang0 (approved):
Everything looks good! Thanks for walking me through it yesterday :)

clarsen-czi:
Everything looks great except for the overall height of the callouts—I believe you have included too much top/bottom padding around the icons and text; the overall height of the callout should be 42px. Get this fixed and I think it's good to go!

tihuan:
Feels like this PR has outdated code compared to the other one, so maybe we just need to clean up this PR after the other is merged? Thank you!

PR #59, opened 2021-09-14
thuang-update-complex-filter

tihuan:
Updated ComplexFilter component per your awesome feedback 🎉 ! Thank you!

Also I experimented with using union callback functions for `onChange` prop, but since TypeScript still can't tell when the callback function argument `options` is expected to be an array or not, it still throws the same error as the ones we're seeing at the call site in Aspen 🤔 

I forgot if you've tried the following solution by telling `useState` the value type is `ComplexFilterValue`, but I think and hope this will work 😆 

```ts
import { ComplexFilterValue } from 'czifui'

function Foo() {
  const [value, setValue] = useState<ComplexFilterValue>([]);

  return ...
}

```


vincent-czi:
Sorry, yeah, I had already tried that option you suggested
```
const [value, setValue] = useState<ComplexFilterValue>([]);
```

Just checked again. Here's the error message that TS outputs:
```
Property 'map' does not exist on type 'DefaultMenuSelectOption | DefaultMenuSelectOption[]'.
  Property 'map' does not exist on type 'DefaultMenuSelectOption'.ts(2339)
```

TS sees `value` after that as being possibly any one of the union (it drops `null` because it gets type narrowed out with a preceding `if (value)` where the error happens). The `DefaultMenuSelectOption[]` part is happy to work with `.map`, but the `DefaultMenuSelectOption` blows up.

I'll keep poking at it as well and let you know if I come up with any new ideas. This seems like such a reasonable way to construct a component, it's remarkable how hard the Type aspect is.

vincent-czi (approved):
This is great! Perfect, and thanks for putting in the docs!

One thing is the changes should necessitate a patch version bump, right? I don't know how your release approach works though, so maybe that's a separate PR.

tihuan:
Oh yes! Definitely need a patch version bump 👍 Thanks for the reminder! Also published to npm now 🎉 

Re `.map()`: Ahh I see, so we need to tell TS `value` is for sure `DefaultMenuSelectOption[]`. What about using type assertion `as`?

Something like: 

```ts
const [value, setValue] = useState<ComplexFilterValue>([]);

if (value) {
 const mappedItems = (value as ComplexFilterValue[]).map(...)
}
```

vincent-czi:
ah yes! That's the trick. Slotting in the type assertion there works correctly. Here's what it looks like now:
```
  const [value, setValue] = useState<ComplexFilterValue>([]);

  useEffect(() => {
    if (value) {
      updateLineageFilter((value as DefaultMenuSelectOption[]).map((d) => d.name));
    }
  }, [updateLineageFilter, value]);
```

I'm still curious about the idea of "two-track Types" like we're seeing here -- it feels like something that should be doable without an assertion -- but this definitely solves my problem. Thanks!

tihuan:
Awesome! So glad that worked 🎉 !

Yeah I just peeked at how Material UI solves this "two-track types" problem with their Autocomplete component, and somehow they're able to change the onChange value type based on whether the component has `multiple === true` or not, so I'll dig a little more into it to see how they link all that together seamlessly 🤩

PR #62, opened 2021-09-21
[155533] Notification Update

mdjmoore:
**To be merged in with #57 to avoid breaking changes**

Shortcut ticket: [155533](https://app.shortcut.com/sci-design-system/story/155533/update-notifications-component-for-sds)

## Description
update the Alert component to Notification as per the reference doc. I've left Alert in storybook and the library so that the component can be depreciated in stages, as mentioned during an [SDS Infra meeting](https://docs.google.com/document/d/1j4tmHiTAm9mpBpXnF5WoXoZEoZlFzW_GkLNq6mGiUOE/edit?usp=sharing). Alert should be phased out only when it's confirmed that it is no longer being used by any client. This allows clients to refactor to the new component on their own time.

Screen Shot 2021-09-21 at 10 34 36 AM

Screen Shot 2021-09-21 at 10 34 43 AM

- [x] Updated component naming to "Notifications" (as in the design documentation).
- [x] Updated container padding > 14px, 14px, 14px, 10px.
- [x] Updated text size to font-body-xs/400
- [x] Update icon for info, warning, and error notifications to the "Alert icon" ($icon-l).
- [x] All icon sizes should be $icon-l.
- [x] Updated icon right margin to spacing-m (10px).
- [] Updated button text size to font-caps-xxxs/600.
- [x] Added notification variation that is dismissable (with close icon, $icon-xs).
- [x] Content area container should be able to be swapped / filled with desired content by whoever is using that component.
- [x] Notification component as well as the various intents can be viewed in storybook.


karenliang0:
Thanks for working on this Meagan!! It's looking good!

Here's a few pieces of feedback I had:

1. The first line of text should be center vertically aligned against the Alert icon regardless of the notification height so in that regard, it should stay static. Currently, it seems like when the notification grows taller from content, the first line of text is moving with it. 

2. The text link should be left aligned with the text! 

3. The close icon should be center vertically aligned against the Alert icon regardless of notification height as well! 

4. The close icon doesn't seem to be 10x10 which is the icon-xs sizing! 

image

I hope this image is useful! Let me know if all of this makes sense and we can always discuss :)

mdjmoore:
Thanks for the feedback @karenliang0. I've addressed it in the most recent commit. One thing to note is that the path for the close icon will be off by 1px (9.01 to be exact), but the svg itself is 10x10. This is true of the icons that I pulled from the SDS ref doc. Let me know if that works or if we need to get some new svgs

src/core/Callout/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { AlertProps } from "@material-ui/lab";
+import React from "react";
+import { ReactComponent as IconAlert } from "../../common/svgs/IconAlert.svg";
+import { ReactComponent as IconClose } from "../../common/svgs/IconClose.svg";
+import { ReactComponent as IconSuccess } from "../../common/svgs/IconSuccess.svg";
+import { StyledCallout } from "./style";
+
+export { AlertProps };
```
tihuan, line 8: Since the component is Callout, I wonder if we should re-export MUI's `AlertProps` as `CalloutProps`?

E.g,

```ts
import { AlertProps as CalloutProps } from "@material-ui/lab";

export { CalloutProps }
```

src/core/Notification/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { AlertProps } from "@material-ui/lab";
+import React from "react";
+import { ReactComponent as IconAlert } from "../../common/svgs/IconAlert.svg";
+import { ReactComponent as IconClose } from "../../common/svgs/IconClose.svg";
+import { ReactComponent as IconSuccess } from "../../common/svgs/IconSuccess.svg";
+import { StyledButton, StyledNotification } from "./style";
+
+export { AlertProps };
```
tihuan, line 8: Same here, maybe `NotificationProps` instead of `AlertProps`, in case `CalloutProps` and `NotificationProps` diverge from `AlertProps` down the road, and we won't need to introduce a breaking change 🙆‍♂️ !

src/core/Notification/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface NotificationProps {
+  sdsAction?: (event: React.SyntheticEvent) => void;
```
tihuan, line 11: Ah instead of `sdsAction` would `onClick` work?

src/core/Notification/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  sdsActionText?: string;
```
tihuan, line 12: maybe `text` would do?

tihuan (commented):
Looks super good! Thanks so much for all the thoughtful stories and code, @mdjmoore 🚀 🙏

src/core/Notification/index.tsx
mdjmoore: I think just `text` might be a bit vague in the scope of the whole component. What about `buttonText`?

src/core/Notification/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  children,
+  intent,
+  onClose,
+  buttonOnClick,
+  buttonText,
+  ...rest
+}: ExposedNotificationProps): JSX.Element => {
+  return (
+    <StyledNotification
+      {...rest}
```
tihuan, line 26: Ah I think `{...rest}` needs to go last in order to allow the callsite to overwrite props!

Example of Maya's [recent fix](https://github.com/chanzuckerberg/sci-components/pull/67/files)

E.g.,

```tsx
<StyledNotification
  action={...}
  ...
  {...rest}
>

```

src/core/Notification/index.tsx
tihuan: Oh yes that's even better 🚀 ! Great idea thank you, Meagan 😄 !

tihuan (approved):
Your code is beautiful 🔥 ✨ !! Thanks so much for the amazing work, @mdjmoore 🚀 !!

clarsen-czi:
This is looking great, Meagan! A few small updates based on conversations this week with the design team:

- We've made top/bottom and left/right padding equal, so it should be `spaceL` all the way around (as opposed to `spaceL` top/bottom and `spaceXl` left/right as it currently is)
- There should be the option to have the Content Module with or without the button (in your current version, the two are tied together)
- We should add in an `autoDismiss` prop which when set to `true`, the Notification will automatically dismiss after 8 seconds
- Can we set the default width to be 360px? This can be adjusted by designs as necessary, but that should be the general starting point

clarsen-czi:
Really looking great, Meagan!

A few small changes:
- Remove the margin-bottom: spaceM from the extraContent block
- Replace the icon on Intent: Info with IconInfo (currently IconAlert)
- For the Live Preview view, can you include 3 views of the component from left to right all with Intent: Info—one with extraContent: false and buttonOnClick: false, one with extraContent: false and buttonOnClick: true, and one with extraContent: true and buttonOnClick: true

PR #63, opened 2021-09-21
tweak styles for minimal input dropdown when open

mmmmmmaya:
### Summary
- **What:** Fix styles for inputdropdown when menu is open
- **Why:** I missed that there was a separate state for this when implementing
- **Env:** http://localhost:6006/?path=/story/inputdropdown--minimal

#### Before: 
before

#### After: 
after

### Checklist
- [X] I merged latest `main`
- [x] I added labels to my PR

### DIFF
--- a/src/core/InputDropdown/styles.ts
+++ b/src/core/InputDropdown/styles.ts
@@ -14,6 +14,7 @@ export interface InputDropdownProps extends Props {
   disabled?: boolean;
   label: string;
   onClick: (event: React.MouseEvent<HTMLElement>) => void;
+  open?: boolean;
   sdsStyle?: "minimal" | "square" | "rounded";
 }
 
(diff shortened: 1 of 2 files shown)


src/core/InputDropdown/styles.ts
```diff
@@ -72,21 +73,35 @@ const minimal = (props: InputDropdownProps): SerializedStyles => {
...
       }
     }
   `;
 };
+
+const isOpen = (props: InputDropdownProps): SerializedStyles => {
+  const palette = getPalette(props);
+  return css`
+    span {
+      color: ${palette?.text?.primary};
```
phoenixAja, line 86: is this color black?

phoenixAja (approved):
looks good, the only thing is that i want to verify we're using the correct colors as reflected in the design (it looked like the carat and text were both using the same color here)

src/core/InputDropdown/styles.ts
mmmmmmaya: the designs actually are not consistent within the design system ref doc. but for [inputdropdown](https://www.figma.com/file/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=1876%3A21139), the user input state shows them both as black (the carat should only be colored on active). For [filters](https://www.figma.com/file/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=3139%3A19517) (which, iirc, should make use of the inputdropdown, so they should be the same).

I'm checking in with SDS about how the inputdropdown component should actually style it in the default application (then aspen will get it for free).

tihuan (approved):
LGTM!! Thanks for improving the component 🎉 !

PR #64, opened 2021-09-22
chore: add Commitizen

tihuan:
This PR follows Commitizen's [documentation](https://github.com/commitizen/cz-cli#husky) to install and add prepare commit message hook to trigger Commitizen command and also added a README section!

From now on when we add a new commit, Commitizen will walk you through a set of questions to properly prepare the right commit message that follows Semantic Versioning 🙆‍♂️ 

NOTE: If you just want to make a quick commit, you can `^ + C` to escape Commitizen and use your existing commit message too!

### DIFF
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+exec </dev/tty && node_modules/.bin/cz --hook || true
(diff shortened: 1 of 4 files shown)


mdjmoore (approved):
(つ▀¯▀)つ

I would flag devs that they need to install commitizen globally and write up/screenshot an example flow for using `git cz` as it should be implemented in our repo and add that to the contribution docs

tihuan:
Thanks so much for the great feedback, @mdjmoore 🙏 ! Somehow I thought Commitizen would get automatically installed when people run `npm i` from now on lol But I think I need to add commitizen to our package.json first 😂 

Will do that now, so people don't need to explicitly install Commitizen globally!

tihuan:
Just updated the PR and will auto merge! Thanks again!

PR #65, opened 2021-09-23
fix(complexfilter): update ComplexFilter value type to be based on mu…

tihuan:
…ltiple prop

Previously onChange's argument is a union of DefaultMenuSelectOption | DefaultMenuSelectOption[] |
null, which doesn't utilize the value of multiple prop to narrow down the possible value. Now we do!

BREAKING CHANGE: type ComplexFilterValue is no longer needed, since now TypeScript is able to infer
the value type based on the value of multiple being true or false

### DIFF
--- a/src/core/ComplexFilter/index.tsx
+++ b/src/core/ComplexFilter/index.tsx
@@ -1,4 +1,4 @@
-import { AutocompleteCloseReason } from "@material-ui/lab";
+import { AutocompleteCloseReason, Value as MUIValue } from "@material-ui/lab";
 import React, { useEffect, useState } from "react";
 import InputDropdown, {
   InputDropdownProps as InputDropdownPropsType,
@@ -13,20 +13,32 @@ export {
   InputDropdown as ComplexFilterInputDropdown,
 };
 
+/**
+ * @deprecated This type is no longer needed and will be removed in the next major version.
+ */
 export type ComplexFilterValue =
   | DefaultMenuSelectOption[]
   | DefaultMenuSelectOption
   | null;
 
-interface ComplexFilterProps {
+// (thuang): Value's type is based on generic type placeholder (T) and Multiple
+// type. If Multiple is true, Value's type is T[] | null.
+// Otherwise, Value's type is T | null.
+// Conditional Type
+// https://www.typescriptlang.org/docs/handbook/2/conditional-types.html
+export type Value<T, Multiple> = Multiple extends undefined | false
+  ? T | null
+  : Array<T> | null;
+
+interface ComplexFilterProps<Multiple> {
   label: string;
   options: DefaultMenuSelectOption[];
-  multiple?: boolean;
+  multiple?: Multiple;
   search?: boolean;
-  onChange: (options: ComplexFilterValue) => void;
+  onChange: (options: Value<DefaultMenuSelectOption, Multiple>) => void;
   MenuSelectProps?: Partial<typeof MenuSelect>;
   InputDropdownProps?: Partial<InputDropdownPropsType>;
-  value?: ComplexFilterValue;
+  value?: Value<DefaultMenuSelectOption, Multiple>;
   style?: React.CSSProperties;
   className?: string;
   PopperComponent?: typeof StyledPopper;
@@ -34,7 +46,9 @@ interface ComplexFilterProps {
   InputDropdownComponent?: typeof InputDropdown;
 }
 
-export default function ComplexFilter({
+export default function ComplexFilter<
+  Multiple extends boolean | undefined = false
+>({
   options,
   label = "",
   multiple = false,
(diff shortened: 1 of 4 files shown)


src/core/ComplexFilter/index.tsx
```diff
@@ -18,23 +18,29 @@ export type ComplexFilterValue =
   | DefaultMenuSelectOption
   | null;
 
-interface ComplexFilterProps {
+export type Value<T, Multiple> = Multiple extends undefined | false
```
tihuan, line 21: This is how we define a conditional type!

vincent-czi: Wow, this really feels like it's at the edge of my understanding. Let me describe it in my own words and see if that's correct:
- a `Value` type is taking two args: `T` and `Multiple`
- The `Multiple extends` is a truthy/falsy check on if the `Multiple` arg "matches" with `undefined | false`
- From there, we have a ternary that sets the resulting Value type based on that match -- truthy becomes `T | null`, while falsy becomes an array of `T` and still `null`.
- Above is because if Multiple "matches" with `undefined | false`, then it was not present in the incoming props, and thus we are in the single case, etc

All that seems to hold up pretty well. The big thing that really freaks me out here is the use of `extends` to do this kind of type checking? I thought `extends` was an imperative/declarative keyword _only_: you used it to inherit from "parent" types, ala OOO. [I can find docs and articles on this usage of `extends`](https://www.typescriptlang.org/docs/handbook/2/objects.html) pretty easily, but I can't find anything at all on this other meaning of `extends` as a "question-asking" form? In this usage it feels like a way of doing `typeof` inside of TS-only code. Did you find any docs on this usage of it?

vincent-czi: Also, maybe this isn't as complicated as it feels to me, but I think a line or two of comments above explaining what's going on here would be good. This seems to be a pretty unusual structure, and I think having a bit of plain English that explains why it exists and what it does would be helpful for those with less TS experience (cough cough me cough)

src/core/ComplexFilter/index.tsx
```diff
@@ -18,23 +18,29 @@ export type ComplexFilterValue =
...
+  ? T | null
+  : Array<T> | null;
+
+interface ComplexFilterProps<Multiple> {
```
vincent-czi, line 25: Okay, so I get that `Multiple` is a type param here and so it has ripple effects throughout the interface, but I don't understand how it actually gets used with the real props for the component definition...?

src/core/ComplexFilter/index.tsx
```diff
@@ -18,23 +18,29 @@ export type ComplexFilterValue =
...
   style?: React.CSSProperties;
   className?: string;
   PopperComponent?: typeof StyledPopper;
   PaperComponent?: typeof StyledPaper;
   InputDropdownComponent?: typeof InputDropdown;
 }
 
-export default function ComplexFilter({
+export default function ComplexFilter<
+  Multiple extends boolean | undefined = false
```
vincent-czi, line 42: ... This part here, where the actual ComplexFilter component is defined. I don't understand how the Multiple filters through the definition and Type declaration and back to the underlying interface. Really curious to have you walk me through it.

vincent-czi (approved):
To the extent that I understand what is going on here, good work. The end result feels like a nice, clean API for the eventual user of czifui. The actual Type internals feel pretty darn mysterious to me, but hopefully that's more about me not getting it than this sort of being the edge of the language's expressiveness...

This would actually rank as a breaking change, which feels pretty wild -- I guess this would necessitate a major version bump down the road? Right now, I'd skip it and just communicate with the teams using czifui, but might be good to think about how you might avoid doing a major version bump -- even if it's just for a couple months and put into a deprecation mode -- for this kind of tweak down the road.

tihuan:
Thanks so much for the detailed feedback, Vincent 🎉 🤩 !

Yes! The use of `extends` as ternary operator is documented under [Conditional Type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html), and MUI uses the same technique to type their Autocomplete component 🙆‍♂️ 

The part where the actual `ComplexFilter<Multiple extends boolean | undefined = false>(...)` component passes a type variable `Multiple` to `ComplexFilterProps<Multiple>` is achieved through [Generic Constraints](https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-constraints). Essentially telling TypeScript that `ComplexFilter` function has a type variable, so use it in `ComplexFilterProps` as needed!

I also used the following example to understand how this all works in VS Code 😆 

```ts
interface Props<Multiple> {
  bar: Multiple;
  // Conditional type
  baz: (value: Multiple extends true ? number : string) => void;
}

// Generic constraints, essentially constraining Multiple to be boolean,
// so Props can evaluate Multiple as boolean above.
function Foo<Multiple extends boolean>({ bar, baz }: Props<Multiple>) {
  return [bar, baz];
}

Foo({ bar: true, baz: (value) => value }); // value is number
Foo({ bar: false, baz: (value) => value }); // value is string
```

Yeah it feels wrong that a type change is a breaking change, but it doesn't feel like `ComplexFilterValue` can coexist with the updated type, since they are contradictory by nature

However, we could publish this type update on `czifui@next` tag first, if that helps give Aspen more time to prepare for the upgrade?

src/core/ComplexFilter/index.tsx
tihuan: Great idea! I added comments to explain the usage being [Conditional Type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) 🚀 !

tihuan:
Sync'd with Vincent offline, we'll publish this on `@latest` as a breaking change for ComplexFilter 🙆‍♂️ 

Thank you!

PR #66, opened 2021-09-24
Button updates

ehoops:
- Style updates to existing buttons.  
- Implements sdsType and sdsStyle as optional props
- Adds more stories to storybook

This is backward compatible, but we should clean up at some point once teams update.

Default state:
Screen Shot 2021-09-23 at 4 55 11 PM

Hover state:
Screen Shot 2021-09-23 at 4 55 15 PM

(Couldn't screenshot the click state, but it's there)

Different states in Storybook:
Screen Shot 2021-09-23 at 5 03 56 PM
Screen Shot 2021-09-23 at 5 03 42 PM

### DIFF
--- a/src/core/Button/index.tsx
+++ b/src/core/Button/index.tsx
@@ -8,21 +8,30 @@ import {
   StyledButton,
 } from "./style";
 
-export interface ButtonProps extends RawButtonProps {
+interface SdsProps {
+  isAllCaps?: boolean;
   isRounded?: boolean;
   sdsStyle?: "minimal" | "rounded" | "square";
   sdsType?: "primary" | "secondary";
 }
+export type ButtonProps = RawButtonProps & SdsProps;
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   (props: ButtonProps, ref): JSX.Element | null => {
     const sdsStyle = props?.sdsStyle;
     const sdsType = props?.sdsType;
+    const isAllCaps = props?.isAllCaps;
 
     if (!sdsStyle || !sdsType) {
       // eslint-disable-next-line no-console
       console.warn(
         "Warning: Buttons without sdsStyle or sdsType props will be deprecated."
+      );
+    }
+    if (isAllCaps && sdsStyle !== "minimal") {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "Warning: isAllCaps is only applied to buttons with sdsStyle='minimal'."
       );
     }
 
--- a/src/core/Button/index.tsx
+++ b/src/core/Button/index.tsx
@@ -15,7 +15,7 @@ export interface ButtonProps extends RawButtonProps {
 }
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
-  (props, ref): JSX.Element | null => {
+  (props: ButtonProps, ref): JSX.Element | null => {
     const sdsStyle = props?.sdsStyle;
     const sdsType = props?.sdsType;
 
--- a/src/core/Button/index.tsx
+++ b/src/core/Button/index.tsx
@@ -1,14 +1,89 @@
 import { ButtonProps as RawButtonProps } from "@material-ui/core";
 import React from "react";
-import { StyledButton } from "./style";
+import {
+  PrimaryMinimalButton,
+  RoundedButton,
+  SecondaryMinimalButton,
+  SquareButton,
+  StyledButton,
+} from "./style";
 
 export interface ButtonProps extends RawButtonProps {
   isRounded?: boolean;
+  sdsStyle?: "minimal" | "rounded" | "square";
+  sdsType?: "primary" | "secondary";
 }
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
-  (props, ref): JSX.Element => {
-    return <StyledButton {...props} ref={ref} />;
+  (props, ref): JSX.Element | null => {
+    const sdsStyle = props?.sdsStyle;
+    const sdsType = props?.sdsType;
+
+    if (!sdsStyle || !sdsType) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "Warning: Buttons without sdsStyle or sdsType props will be deprecated."
+      );
+    }
+
+    switch (true) {
+      case sdsStyle === "rounded" && sdsType === "primary":
+        return (
+          <RoundedButton
+            color="primary"
+            ref={ref}
+            variant="contained"
+            {...props}
+          />
+        );
+      case sdsStyle === "rounded" && sdsType === "secondary":
+        return (
+          <RoundedButton
+            color="primary"
+            ref={ref}
+            variant="outlined"
+            {...props}
+          />
+        );
+      case sdsStyle === "square" && sdsType === "primary":
+        return (
+          <SquareButton
+            color="primary"
+            ref={ref}
+            variant="contained"
+            {...props}
+          />
+        );
+      case sdsStyle === "square" && sdsType === "secondary":
+        return (
+          <SquareButton
+            color="primary"
+            ref={ref}
+            variant="outlined"
+            {...props}
+          />
+        );
+      case sdsStyle === "minimal" && sdsType === "primary":
+        return (
+          <PrimaryMinimalButton
+            color="primary"
+            ref={ref}
+            variant="text"
+            {...props}
+          />
+        );
+      case sdsStyle === "minimal" && sdsType === "secondary":
+        return (
+          <SecondaryMinimalButton
+            color="default"
+            ref={ref}
+            variant="text"
+            {...props}
+          />
+        );
+      default:
+        return <StyledButton {...props} ref={ref} />;
+    }
   }
 );
 
--- a/src/core/Button/index.tsx
+++ b/src/core/Button/index.tsx
@@ -20,7 +20,6 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   (props: ButtonProps, ref): JSX.Element | null => {
     const sdsStyle = props?.sdsStyle;
     const sdsType = props?.sdsType;
-    const isAllCaps = props?.isAllCaps;
 
     if (!sdsStyle || !sdsType) {
       // eslint-disable-next-line no-console
@@ -28,12 +27,18 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         "Warning: Buttons without sdsStyle or sdsType props will be deprecated."
       );
     }
-    if (isAllCaps && sdsStyle !== "minimal") {
+
+    if (typeof props?.isAllCaps === "boolean" && sdsStyle !== "minimal") {
       // eslint-disable-next-line no-console
       console.warn(
         "Warning: isAllCaps is only applied to buttons with sdsStyle='minimal'."
       );
     }
+
+    // isAllCaps is only used for the Minimal Button.  It defaults to true.
+    const isAllCaps =
+      typeof props?.isAllCaps === "boolean" ? props?.isAllCaps : true;
+    const propsWithDefault = { ...props, isAllCaps };
 
     switch (true) {
       case sdsStyle === "rounded" && sdsType === "primary":
@@ -42,7 +47,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
             color="primary"
             ref={ref}
             variant="contained"
-            {...props}
+            {...propsWithDefault}
           />
         );
       case sdsStyle === "rounded" && sdsType === "secondary":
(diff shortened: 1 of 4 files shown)


happyimadesignr:
Hey @ehoops are all the dropshadows gone from the buttons now?

The text button should be all caps. I know we changed some things to title case for accessibility but in this case it's important to differentiate a text button from a regular link.

@clarsen-czi can you help review this PR for visual details?

src/core/Button/style.ts
```diff
@@ -1,11 +1,93 @@
...
+      min-width: 120px;
+      height: 34px;
+      &:hover {
+        color: white;
+        background-color: ${colors?.primary[500]};
+        box-shadow: none;
+      }
+      &:active {
+        color: white;
+        background-color: ${colors?.primary[600]};
```
clarsen-czi, line 22: We should also have `box-shadow: none;` on `:active`

ehoops:
> Hey @ehoops are all the dropshadows gone from the buttons now?
> 
> The text button should be all caps. I know we changed some things to title case for accessibility but in this case it's important to differentiate a text button from a regular link.
> 
> @clarsen-czi can you help review this PR for visual details?

@happyimadesignr Yes, the drop shadows are gone - this was intentional based on the ticket description: https://app.shortcut.com/sci-design-system/story/152362/update-existing-buttons
I'm happy to put them back if we want them (or maybe that was only for a specific state?)

The text of the button isn't currently controlled by the component.  Whoever uses the button puts in their own text/icons.  If we want to control this, we can.  It would make it hard for people to override though.

clarsen-czi (changes requested):
Looks great Erin! Here are just a few more changes in addition to the comment I left on the code about removing drop shadows from the active state:

- Echoing Jenn's comment about using `fontCapsXxxs` for the Minimal buttons
- The Minimal button should have top and bottom padding of `spaceXxs` (for a total height of '24px' since `fontCapsXxxs` has a line height of 16px) and no left/right padding
- Both Rounded and Square Secondary buttons should have no border on `:hover` or `:active`
- The background color on the `:disabled` Primary and Secondary buttons should be `gray300` which is `#cccccc` (inspect is showing it `#0000001f`); not sure why a transparent value is being used for `gray300` instead of our defined color, so can you check on that?
- I'm sure you're aware of this, but there should be the option for icons for Primary and Secondary Round and Square buttons, as well as Primary Minimal button.
- We no longer have a button with Tooltip, so please remove that from the repo

ehoops:
I know there are places in Rare As One where we use text buttons without all caps font, so we would need to override in order to use this if we change it to making the text all caps by default.

My biggest gripe with it goes back to accessibility.  Using all caps by default goes against our attempt to make the easiest thing accessible.  On the other hand, it is more accessible to use css to capitalize font rather than just typing in all caps.  If we do this ourselves, we would be able to do it with css by default and reduce the chance of a product engineer not knowing that.

Thoughts @tihuan @j-x-han @mmmmmmaya ?

src/core/Button/index.tsx
```diff
@@ -1,14 +1,89 @@
...
-  (props, ref): JSX.Element => {
-    return <StyledButton {...props} ref={ref} />;
+  (props: ButtonProps, ref): JSX.Element | null => {
+    const sdsStyle = props?.sdsStyle;
+    const sdsType = props?.sdsType;
+
+    if (!sdsStyle || !sdsType) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "Warning: Buttons without sdsStyle or sdsType props will be deprecated."
```
tihuan, line 25: love the warning!!

tihuan (approved):
LGTM! Thanks so much for enhancing all the buttons 🎉 !!

Yeah I also think that leaving all caps mixin as an opt-in by the project team is more flexible and aligned with a11y best practices. 

If we think it's a frequent use case, we could also add a new prop `allCaps`?

clarsen-czi:
Thanks for the changes, Erin!

Everything looks great except for a few small things I still see:
- Can you remove the left and right padding on the Minimal button
- Can you reduce the top and bottom padding on the Minimal button to spaceXxs (4px)

A few questions I have:
- I like Timmy's idea of including an `isAllCaps` prop to make it easy to toggle between the two font styles. If we go that route, are we able to just swap in the `fontCapsXxxs` mixin for the `fontBodyXs` one? `fontCapsXxxs` adjusts the font size and letter spacing in addition to making the font all caps, whereas if we just use CSS `text-transform: uppercase` but keep everything else the same, the text will appear too large on the button and the letters will be too close together making it harder to read.
- By not having variants with icons for each button type, are we going to run into issues with product teams adding incorrect icon sizes and padding around them? I guess I just want to make sure this is standardized on each button type so they always appear in the same location and at the same size.

clarsen-czi (commented):
Looking good to me—see additional comments/questions in my comment above (sorry, I thought adding those comments yesterday was enough to submit my review, didn't realize I needed to come to this tab to do it)

happyimadesignr:
Hi @ehoops @clarsen-czi @Janeece @tihuan and team,

**I would suggest us taking a stance and making text buttons all caps in SDS because:**
- Text buttons should have a specific look to communicate their function. The visual style of elements should communicate an expectation to the user of what will happen.
- We don't want text buttons to be confused with links.
- I would not be comfortable recommending people use title case for text buttons because I don't think we would be giving them the best advice. (See more below, I would recommend a different design for that RAO button)
- I don't think it's actually breaking an accessibility rule because it's only 1-2 words and the recommendations I've seen are regarding caps is about reading many words in sentences and paragraphs. Let me know if there's something I'm missing.

@ehoops:
- Thanks for removing the shadows, I'm sooo excited those will be gone!!! I'm sure @Janeece is also stoked!
- Would love to hear more about your accessibility concern. What specific accessibility issue is your concern in this case? Is it about the legibility of all caps?
- I see one of the buttons you're referring to in RAO, "Looking for a specific guide? Let us know." is a button because it opens a dialog. That one wouldn't qualify in this definition for a text button. Text buttons should mostly be one word, maybe 2, if you have 3 you're really really stretching it, please don't use 3 words! (@clarsen-czi can you help make sure these rules about text buttons are captured if not already!) 
- For that particular button, I would actually recommend a different design rather than making those 2 sentences into a button. This goes back to the principle that the visual style of elements should communicate an expectation to the user of what will happen. In this case, it's not clear to the user based on the visuals that this link is opening a dialog, I expected it to open a page. This would be a usability issue and I would also think would be an accessibility concern, what do you think?

Also, @ehoops @tihuan, I see the default MUI button are already all caps, curious why there would be an implementation challenge to do that?

Thanks so much!!

ehoops:
> Thanks for the changes, Erin!
> 
> Everything looks great except for a few small things I still see:
> 
> * Can you remove the left and right padding on the Minimal button
> * Can you reduce the top and bottom padding on the Minimal button to spaceXxs (4px)
> 
> A few questions I have:
> 
> * I like Timmy's idea of including an `isAllCaps` prop to make it easy to toggle between the two font styles. If we go that route, are we able to just swap in the `fontCapsXxxs` mixin for the `fontBodyXs` one? `fontCapsXxxs` adjusts the font size and letter spacing in addition to making the font all caps, whereas if we just use CSS `text-transform: uppercase` but keep everything else the same, the text will appear too large on the button and the letters will be too close together making it harder to read.
> * By not having variants with icons for each button type, are we going to run into issues with product teams adding incorrect icon sizes and padding around them? I guess I just want to make sure this is standardized on each button type so they always appear in the same location and at the same size.

√ for the padding changes.  I had a typo, so they weren't actually applied.  They will be now.
√ I added an isAllCaps prop to apply fontCapsXxxs.  This won't necessarily stop people from passing in other sized caps (we can't do that), but hopefully it will make it easy to follow the designs.

For the icons: The buttons already have a built in prop from MUI called 'startIcon' which places the icon on the correct part of the button with built in spacing.  We use this in the storybook examples, so 🤞 engineers see this and also make use of the prop.  I don't think there's much more we can add here.  While it's possible to restrict buttons to only contain text, it's not standard.  The design system can/should make the easiest path the most right, but if someone really wants to put in more effort they can always use it incorrectly.  We also run the risk of people just not using the design system if we make it too complicated/non-standard.

ehoops:
Please lmk if this is too small to see.  I figured you would be zooming anyway.

Screen Shot 2021-09-30 at 10 55 09 AM

vincent-czi:
Regarding making text buttons all caps across SDS, I wanted to point out that we currently use quite a few not-all-caps buttons in Aspen. Here are some examples:

image

image

From the conversation above, it sounds like it wouldn't be enforced as always all caps, but czifui `Button` is definitely out in the wild, not being used with all caps currently.

ehoops:
@happyimadesignr 

> * I see one of the buttons you're referring to in RAO, "Looking for a specific guide? Let us know." is a button because it opens a dialog. That one wouldn't qualify in this definition for a text button. Text buttons should mostly be one word, maybe 2, if you have 3 you're really really stretching it, please don't use 3 words! (@clarsen-czi can you help make sure these rules about text buttons are captured if not already!)ƒ
> * For that particular button, I would actually recommend a different design rather than making those 2 sentences into a button. This goes back to the principle that the visual style of elements should communicate an expectation to the user of what will happen. In this case, it's not clear to the user based on the visuals that this link is opening a dialog, I expected it to open a page. This would be a usability issue and I would also think would be an accessibility concern, what do you think?

The alternative here is that RAO eng has to create our own button that is text and not use the design system.  I think it's not great to not accommodate the designs because the alternative is not getting the other built in benefits of the SDS.  If this is more for designers to adhere to the recommendation, that's fine, but once the engineers are told what to build our process is usually "use what's available if it works for the design and if not, build it".

>Also, @ehoops @tihuan, I see the default MUI button are already all caps, curious why there would be an implementation challenge to do that?

This is quite weird.  I see what you mean on the MUI documentation, but we aren't doing anything to fight this.  I'm not sure why they aren't all caps by default.  I don't think we should make this restrictive when different products have different use cases.  The option is there for the `isAllCaps` prop now.

Janeece:
Re @vincent-czi call out: yeah the majority of buttons (rounded and square styles) should remain title cases. The all caps design recommendation is for minimal buttons only, specifically because these buttons do not have a visible container.

I'm also wondering if part of the issue tracks back to the need for "buttons that look like links" and "links that look like buttons" for better accessibility? If so, then yeah, we'd need a button that looks like text links, but I it would be different than the all caps minimal button.

ehoops:
@happyimadesignr @Janeece Another option is to have caps be the default, but an engineer could use `isAllCaps=false` to easily override this if/when needed

vincent-czi:
> The design system can/should make the easiest path the most right, but if someone really wants to put in more effort they can always use it incorrectly.

I think there's a real tension here. Usually this work is not done by a single person who is deciding that the trade-off of "incorrect usage" is worth "more effort". If it were, I think a firm hand in design decisions at the level of SDS would be pretty reasonable, because the trade-offs would be fully felt by the person making them. But usually the person doing the design work and the person doing the engineering implementation are different people. Making the choice to design UX that pushes against intended SDS usage is being done by a different person than the one actually writing the code. 

Like @ehoops just said above
> once the engineers are told what to build our process is usually "use what's available if it works for the design and if not, build it".

I think a lot of the choice to make strong, opinionated design choices in SDS comes down to what kind of buy-in you have from designers on the downstream products. For Aspen, I don't think this will be an issue since @Janeece is an active part of the SDS project, but it can be a real difficulty in those cases when a designer is not keeping SDS at the front of mind while working.

tihuan:
Yeah I was also thinking about the button/link look alike situation could possibly be the crux of the problem?

If so, it seems like many UX design sources advocate for links to look like links and buttons to look like buttons

https://marcysutton.com/links-vs-buttons-in-modern-web-applications
https://www.freshconsulting.com/insights/blog/uiux-principle-19-buttons-should-look-like-buttons-links-should-look-like-links/

Side note: The following two sources recommend using CSS `uppercase` to transform text to all caps, instead of actual text being all caps, to prevent screen readers from reading each letter. And also one step further in the a11y game by adding option to allow users to disable all caps if they want to

https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Capitalization 
https://hilo.hawaii.edu/help/accessibility/#capitalization

ehoops:
I am all for buttons looking like buttons and links looking like links.  However, I have never worked on a product that was designed like this all the way through.  I don't want to be too optimistic in expecting future/thrid party designers to follow these best practices, especially when most products don't follow them.

Janeece:
I think this discussion needs a meeting! Seems too complicated for commenting only.

ehoops:
Text buttons are now caps by default.  This can be overwritten using `isAllCaps = false`
Screen Shot 2021-09-30 at 2 05 56 PM

tihuan (approved):
LGTM! Thanks for the quick change, @ehoops 🎉 !!

clarsen-czi (approved):
Thanks for the quick turnaround on these changes and clarification on how the icon buttons will work, @ehoops!! Everything looks great!

PR #68, opened 2021-09-30
chore: semantic release

tihuan:
1. Use `npx semantic-release-cli setup` to add `NPM_TOKEN` and `GITHUB_TOKEN` for Github Action "Release" ([example](https://github.com/chanzuckerberg/sci-components/runs/3750647844?check_suite_focus=true)), so the action can generate release notes, tag related PRs, and publish to `npm` package automatically

2. Create Github Action "Release" that's triggered every time `main` and `next` get new commit, which will run Semantic Release to to all the things above

3. Looks like our `package.json`'s version got changed from `0.0.56` to `0.0.0-development`, so not sure what will be the next npm version after this PR merges to `main` and Semantic Release does another release. But I think Semantic Release will determine the right version for us, so we don't need to think about it 😆 ?

4. Add `commitlint` Github Action to lint commit message format!

PTAL thank you!


tfrcarvalho:
Couple of questions:
* We should clarify the version no? I think we should start with 0.1.0 at least. Let me know if you want me to help take a look.
* Did you gave it any thought of avoiding 1.0.0 for a while?

tihuan:
1. Oh now thinking about it, I can do it dry run on this PR to see what version Semantic Release will bump!
1. Ah yeah I added some thoughts in the [original comment thread](https://docs.google.com/document/d/1j4tmHiTAm9mpBpXnF5WoXoZEoZlFzW_GkLNq6mGiUOE/edit?disco=AAAAPtHeMUg)!   

Overall, I'm thinking that since products already rely on SDS library in prod, it might be easier for them to catch breaking changes by observing the major version bump, instead of paying attention to every changelog. Thus I'm leaning more towards letting Semantic Release does its thing and bump major version whenever it wants

And I also have a feeling that `czifui` is a temp name, so once the library gets more stable in the next 6 months, we might be able to use an official package name and debut the "real" 1.0.0 then?

vincent-czi:
> Overall, I'm thinking that since products already rely on SDS library in prod, it might be easier for them to catch breaking changes by observing the major version bump, instead of paying attention to every changelog. Thus I'm leaning more towards letting Semantic Release do its thing and bump major version whenever it wants

While I would hope this would work out, you never know when something will be missed and a minor/patch will inadvertently introduce a breaking change (or maybe not _breaking_, but enough visual discrepancy it's a problem). Whenever I need to do a version bump of czifui, I plan to look over the changelog then use that as a guide of where to focus during some light regression testing.

I don't know how smart `semantic-release-cli` is -- maybe it will catch everything that could cause discrepancies and bump major version accordingly -- but considering how dependent Aspen is on czifui, it would certainly take some time for me to trust it without doing any kind of regression testing on my own.

tihuan:
Yeah Semantic Release solely bases its versioning on the commit messages types, so if we provide the wrong commit message type, Semantic Release will bump the wrong version. So we'll need to catch that at PR review time 👀 

Currently `main` already has a Commitizen hook to help authors craft the right commit type, so that should help limit the issue. And on top of that, this PR also adds commit message format check, so if future PRs don't follow conventional commit format, it will flag it!

Visual discrepancy will be the next big topic we need to cover in Infra work, since it applies to all version changes 🤔

mdjmoore (approved):
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

tfrcarvalho (approved):

tfrcarvalho:
> And I also have a feeling that `czifui` is a temp name, so once the library gets more stable in the next 6 months, we might be able to use an official package name and debut the "real" 1.0.0 then?

sounds good to me... Not the most attractive name. :)

tihuan:
> > And I also have a feeling that `czifui` is a temp name, so once the library gets more stable in the next 6 months, we might be able to use an official package name and debut the "real" 1.0.0 then?
> 
> sounds good to me... Not the most attractive name. :)

Oh gosh I definitely regret picking that name every day 😂 So hard to type and say too 😭 🤣

tihuan:
Thanks everyone! I think I'll merge this on Monday, since releasing on Friday is too much 😆

PR #73, opened 2021-10-05
docs: update contribution docs with `next` branch merge strategy

tihuan:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "czifui",
-  "version": "1.0.1",
+  "version": "0.0.0-development",
```
tihuan, line 3: turned out that Semantic Release ignores this and relies on commit history to determine the next version 😆 Reverting this back to `0.0.0-development` to avoid confusion

PR #74, opened 2021-10-05
fix(loadingindicator): update SVG import to use relative path

tihuan:
It seems like using relative path should work by looking at the build artifact 🤔 But we might need to investigate more as to why we can't use absolute import 🔎 

BEFORE:

Screen Shot 2021-10-05 at 4 03 48 PM

AFTER:
Screen Shot 2021-10-05 at 4 04 11 PM


mmmmmmaya (approved):
🤞

PR #76, opened 2021-10-12
Maya/dropdown

mmmmmmaya:
### Summary
- **What:** Add a dropdown component
- **Why:** Aspen needs one for usher work, so want to have something more generic available for everyone to use

### Testing
http://localhost:6006/?path=/story/dropdown--single-select

### Notes
At this point, this is basically a direct copy of ComplexFilter, without the involvement of chips. However, since ComplexFilter has a lot of spec'd out design work to add different kinds of filters (sliders/ranges, inline multi select, etc), I think it makes sense to snapshot this Dropdown as its own component. That way, as we modifying ComplexFilter over time to accommodate varying filter styles, Dropdown can remain static and trustworthy for other purposes. The first step in that evolution probably involves replacing the existing ComplexFilter with a Dropdown more or less directly, but I didn't have time to mess around with state management (for example, if ComplexFilter renders a Dropdown, how would I update the Dropdown state by closing out a Chip?), so I am leaving it as-is for now.

### Checklist
- [x] I merged latest `main`
- [x] I added labels to my PR

### DIFF
--- a/src/core/Dropdown/index.stories.tsx
+++ b/src/core/Dropdown/index.stories.tsx
@@ -1,8 +1,6 @@
 import { Args, Story } from "@storybook/react";
-import React, { useState } from "react";
+import React from "react";
 import { noop } from "src/common/utils";
-import Button from "../Button";
-import { DefaultMenuSelectOption } from "../MenuSelect";
 import Dropdown from "./index";
 
 const Demo = (props: Args): JSX.Element => {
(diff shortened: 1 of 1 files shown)


tihuan (approved):
Ooh this is a great idea 💡 ! Thanks so much for adding comments and suggesting a path forward, @mmmmmmaya 🙏 LGTM!!

PR #78, opened 2021-10-14
Add getSpaces selector.

ehoops:
In order to ease some start-up confusion, this adds a `getSpaces` selector.  It is backward compatible and does not remove the `getSpacings` selector.


mdjmoore (approved):
(つ▀¯▀)つ

amazing!!

PR #79, opened 2021-10-14
test(global): add testing

mdjmoore:
## Shortcut stories:
[160185](https://app.shortcut.com/sci-design-system/story/160185/spike-testing-infra-scoping), [164430](https://app.shortcut.com/sci-design-system/story/164430/contribute-to-code-guidelines)

Implements Jest integration tests and snapshot testing to the repo. Also updates contribution docs with new testing info.

Tests run locally per commit as well as on push with Github Actions. Failing tests at either end will block merging. 

Snapshot tests will render a DOM Tree for each component and then will capture new snapshots each time the test suite is run. If they are different (due to changes to the component itself, errors when rendering or style changes that update the generated css class), a warning is generated. 

Test files should contain: snapshot tests, tests to check if the component renders and tests to verify the component handle prop changes

### DIFF
--- a/.svgrrc
+++ b/.svgrrc
@@ -1,6 +1,8 @@
+
 {
+	"__comment": "fillcontrast uses all lowercase to prevent react errors from cluttering the console and test outputs. Props that are in camelCase and passed down by {...rest} are parsed by react and lead to an Unknown Prop Warning. By using lowercase, we pass it along as a custom html attribute and the error is removed",
 	"replaceAttrValues": {
 	  "#000": "{props.fill}",
-	  "#c00": "{props.fillcontrast}",
+	  "#c00": "{props.fillcontrast}" 
 	}
 }


tihuan:
PR looks really good so far! Thanks so much for adding all this, @mdjmoore 🎉 !

For my own education, why did we need to change the prop from `fillContrast` to `fillcontrast`? Thank you!

mdjmoore:
> PR looks really good so far! Thanks so much for adding all this, @mdjmoore 🎉 !
> 
> For my own education, why did we need to change the prop from `fillContrast` to `fillcontrast`? Thank you!

props with camel case are passed onto DOM elements. In this case, `fillContrast` isn't a recognized attribute and causes errors in tests and the console, but we use it internally for recolouring svgs. Making it all lowercase stops it from being passed down to the svg element and removes the console errors

.svgrrc
```diff
@@ -1,6 +1,6 @@
 {
 	"replaceAttrValues": {
 	  "#000": "{props.fill}",
-	  "#c00": "{props.fillContrast}",
+	  "#c00": "{props.fillcontrast}",
```
tihuan, line 4: Can we please add a comment here for future reference on why we use all lowercase 🙏 ? In case we need to add more props or someone wants to refactor this back to camelCase 😆  Thank you!

tihuan (approved):
LGTM!!! Amazing work again, @mdjmoore 🎉

.svgrrc
mdjmoore: will do!

PR #89, opened 2021-11-15
feat(style-dictionary): generate css and scss variables

mdjmoore:
_labelled as a feat because it's a major upgrade and should trigger a build and publish_

This implements [style-dictionary](https://amzn.github.io/style-dictionary/#/) into czifui. Currently this generates css and scss stylesheets using czifui bases. 

Contribution docs have been updated to explain how to use style-dictionary.

The generated stylesheets have been tested locally and can be imported into js files to make the variables available. Currently, because of the way the code is compiled, the stylesheets need to be imported from the `dist` folder. This can be changed, but has bigger consequences that I think outweigh having an extra folder in the import statement currently.

@chanzuckerberg/sds-design please look over the css and scss variables and let me know if this is the syntax and style we want to use


README.md
```diff
@@ -161,6 +161,25 @@ To use the default theme, please do the following:
...
+
+These stylesheets can be imported into the index.js file of your application to make use of czifui's standard styles in projects that use SCSS or CSS modules to style the front end:
+
+```js
+// example using create-react-app structure
+// index.js
+
+import React from "react";
+import ReactDOM from "react-dom";
+import "czifui/dist/variables.css";
```
tihuan, line 178: I was thinking about adding `scss` example too, especially since the export file has `_` prefix!

```suggestion
// If you need css
import "czifui/dist/variables.css";
// If you need scss
import "czifui/dist/_variables.scss";
```

tihuan (approved):
This is SO exciting!! Thanks so much for adding this, @mdjmoore 🤩 👏 ✨ 

Just two tiny questions, otherwise LGTM 🚀 !

src/common/styles-dictionary/css/variables.css
```diff
@@ -0,0 +1,230 @@
+/**
+ * Do not edit directly
+ * Generated on Mon, 15 Nov 2021 22:54:50 GMT
+ */
+
+:root {
+  --sds-font-body-xxxs-400-font-family: "Open Sans", "OpenSans-Regular";
```
mdjmoore, line 7: I was hoping to make the whole rule for each font token into a single variable that held all the values, but letter-spacing isn't part of the css `font` shorthand, so each property gets its own variable

src/common/styles-dictionary/scss/_variables.scss
```diff
@@ -0,0 +1,226 @@
...
+$sds-font-body-xxxs-400-line-height: 16px;
+$sds-font-body-xxxs-400-font-weight: 400;
+$sds-font-body-xxxs-400-color: black;
+$sds-font-body-xxxs-600-font-family: 'Open Sans', 'OpenSans-SemiBold';
+$sds-font-body-xxxs-600-font-size: 11px;
+$sds-font-body-xxxs-600-letter-spacing: 0.3px;
+$sds-font-body-xxxs-600-line-height: 16px;
+$sds-font-body-xxxs-600-font-weight: 600;
+$sds-font-body-xxxs-600-color: black;
+$sds-font-body-xxxs-button-font-family: 'Open Sans', 'OpenSans-Regular';
```
clarsen-czi, line 17: I'm wondering what this button-specific font is since the fonts on our buttons should tie back to existing text styles. Is there something we need to update on our end?

clarsen-czi (commented):
Everything looks good to me except for the one question I have about the button-specific font

src/common/styles-dictionary/scss/_variables.scss
mdjmoore: nope, you're correct. it's the same font. These values were taken from the defaultTheme.ts file. Looking into it, I don't see the button font being used anywhere. I'll do a quick check with the clients before removing it

mdjmoore: it's part of the MUI default theme we created for czifui so I'm removing it from the styles-dictionary but leaving it in defaultTheme.ts

PR #91, opened 2021-11-19
fix(button): fix keyboard navigation for Button. Hover state on focus…

ehoops:
… and add default focus borders

MUI removes the default focus borders for buttons, so keyboard users aren't able to navigate. This
PR applies our :hover styles to :focus as well so they will be visible for keyboard users. It also
puts back the focus borders - these are only visible to keyboard users.

This is the focus state for secondary button (same as hover state):
Screen Shot 2021-11-19 at 8 22 39 AM

This is what it looks like when someone uses a keyboard to get to a minimal button - now they can see they are on a button :
Screen Shot 2021-11-19 at 8 23 36 AM

### DIFF
--- a/src/core/Button/style.ts
+++ b/src/core/Button/style.ts
@@ -24,11 +24,15 @@ const ButtonBase = styled(Button, {
       padding: ${spacings?.xs}px ${spacings?.l}px;
       min-width: 120px;
       height: 34px;
-      &:hover {
+      &:hover, &:focus {
         color: white;
         background-color: ${colors?.primary[500]};
         border: none;
         box-shadow: none;
+      }
+      &:focus {
+        outline: 5px auto Highlight;
+        outline: 5px auto -webkit-focus-ring-color;
       }
       &:active {
         color: white;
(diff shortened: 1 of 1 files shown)


src/core/Button/style.ts
```diff
@@ -24,12 +24,16 @@ const ButtonBase = styled(Button, {
...
       height: 34px;
-      &:hover {
+      &:hover, &:focus {
         color: white;
         background-color: ${colors?.primary[500]};
         border: none;
         box-shadow: none;
       }
+      &:focus {
+        outline: 5px auto Highlight;
```
mdjmoore, line 34: did not know about `Highlight`, very cool!

mdjmoore (approved):
s( ^ ‿ ^)-b

src/core/Button/style.ts
tihuan: ooh same!! Really cool 💡 👏

ehoops: me neither, someone from the internet use it 😅

PR #92, opened 2021-11-19
feat(menuselect): add onInputChange prop to MenuSelect

lvreynoso:
This PR adds an `onInputChange` prop of type `(event: React.SyntheticEvent) => void` that is passed to the `onChange` prop of the child `StyledInputBase` component. This allows anyone using the `MenuSelect` component to pass a callback function that fires whenever a user changes input in the search box.

This functionality is needed by Aspen - we want a dropdown list of locations, but there are >20,000 possibilities. We need this functionality to limit what options are passed to the component based on what the user is searching for.

### DIFF
--- a/src/core/MenuSelect/index.tsx
+++ b/src/core/MenuSelect/index.tsx
@@ -5,6 +5,7 @@ import {
   AutocompleteRenderOptionState,
 } from "@material-ui/lab";
 import React from "react";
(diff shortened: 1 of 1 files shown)


tihuan (approved):
Thanks so much for adding this, @lvreynoso 🙏 Really appreciate the great PR description too! 🤩 👏

LGTM!!

PR #94, opened 2021-12-03
fix(styles-dictionary): refactor typography 

mdjmoore:
**ticket:** [here](https://app.shortcut.com/sci-design-system/story/175588/refactor-typography-variables-in-styles-dictionary)

refactor typography variables to use the font short hand. This will ensure end users will use the appropriate grouped styles when using our design system. Also add global font variables for color, letter-spacing and font family for wider usage and because the `font` shorthand doesn't support color or letter-spacing


tihuan (approved):
LGTM! Thanks for streamlining the values, @mdjmoore 🎉 👏 🤩 !

PR #97, opened 2021-12-16
fix(complexfilter): fix ComplexFilter show virtual keyboard on hidden…

tihuan:
… input

ComplexFilter and MenuSelect has autoFocus on input field, even when search is not enabled. This
causes mobile devices to show virtual keyboard on the hidden input field unnecessarily


mdjmoore (approved):
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

PR #98, opened 2021-12-17
fix(complexfilter): new approach, since readOnly causes weird page ju…

tihuan:
…mp issue on mobile

New approach is using attribute inputMode and css caret-color to hide mobile keyboard

demo

### DIFF
--- a/src/core/ComplexFilter/index.tsx
+++ b/src/core/ComplexFilter/index.tsx
@@ -1,3 +1,4 @@
+import { ClickAwayListener } from "@material-ui/core";
 import { AutocompleteCloseReason, Value as MUIValue } from "@material-ui/lab";
 import React, { useEffect, useState } from "react";
 import { Value } from "../Dropdown";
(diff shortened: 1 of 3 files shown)


src/core/ComplexFilter/index.tsx
```diff
@@ -88,25 +89,29 @@ export default function ComplexFilter<
...
-              undefined,
-              undefined
-            >
-          }
-          onChange={handleChange}
-          disableCloseOnSelect={multiple}
-          options={options}
-          {...MenuSelectProps}
-        />
+        <ClickAwayListener onClickAway={handleClose}>
```
codemonkey800, line 92: amazing

codemonkey800 (approved):
This looks good to me! Just tested this on napari hub and I'm no longer seeing the cursor or the keyboard. awesome work! 😄

mdjmoore (approved):

src/core/ComplexFilter/index.tsx
tihuan: Thank you 😄 ! Actually found this from v5's Autocomplete Github Picker [example](https://mui.com/components/autocomplete/#githubs-picker), which their [v4 example](https://v4.mui.com/components/autocomplete/#githubs-picker) doesn't use and has the bug of clicking the click target won't dismiss the dropdown overlay 💡 

So using `<ClickAwayListener />` here is a bonus fix for that particular bug!

demo

PR #99, opened 2021-12-21
fix(complexfilter): fix update filter state on close

codemonkey800:
## Description

Fixes state update on click away listener close. This works by storing the pending value state in a ref so that the `handleClose` function can access the updated state.

This was the simplest solution I can think of, but let me know if there's another approach I should consider 😄 

### Before

https://user-images.githubusercontent.com/2176050/147010419-4be3c20f-3ee8-433a-b33b-4b3575c9e5ef.mp4

### After

https://user-images.githubusercontent.com/2176050/147010471-0fce4697-a25e-4305-b7c6-871b29d82619.mp4

### DIFF
--- a/src/core/ComplexFilter/index.tsx
+++ b/src/core/ComplexFilter/index.tsx
@@ -1,6 +1,6 @@
 import { ClickAwayListener } from "@material-ui/core";
 import { AutocompleteCloseReason, Value as MUIValue } from "@material-ui/lab";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { Value } from "../Dropdown";
 import InputDropdown, {
   InputDropdownProps as InputDropdownPropsType,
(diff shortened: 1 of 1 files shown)


src/core/ComplexFilter/index.tsx
```diff
@@ -126,7 +130,15 @@ export default function ComplexFilter<
 
   function handleClose() {
     if (multiple) {
-      setValue(pendingValue as Value<DefaultMenuSelectOption, Multiple>);
+      // Set pending value from ref on close. This needs to happen through a ref
```
tihuan, line 133: I just created another PR for the fix here 🙏 

https://github.com/chanzuckerberg/sci-components/pull/100

I think the closure is working as expected here, since when we select an option, React updates `pendingValue` as expected. (I added console.log('pendingValue', pendingValue) before the return statement to verify) 

And the reason why  `pendingValue` is `[]` by the time `handleClose()` is invoked is because `handleClick()` explicitly calls `setPendingValue(value)` BEFORE `handleClose` is invoked, so the solution in my PR is to remove that explicit call in `handleClick()`

PTAL thank you!

codemonkey800:
Closing in favor of #100

PR #103, opened 2022-01-07
fix(iconbutton): iconButton now forwards ref

tihuan:
Since functional components don't have `ref` prop, we need to use `forwardRef` to pass `props.ref` to the target component manually!

Thanks again for spotting this, @codemonkey800 🙏 

Context: https://docs.google.com/document/d/1p7y8qptmw4rrURSEjV7ApAqAzBNMIhQKjn7Obz69dJo/edit#bookmark=id.kx5wxx9qmisj


j-x-han (approved):
Thanks for the fix! 🎉

PR #105, opened 2022-01-14
refactor(tooltip): add warning for using tooltip with a disabled comp…

tihuan:
…onent. Also add forwardRef


j-x-han (approved):
LGTM! Thank you for adding this so quickly 🎉

tihuan:
Yay! Go team 🤩 Thank you for the great suggestion and quick review too 👏 !

PR #106, opened 2022-01-25
feat(dialog): dialog

tihuan:
This PR implements Dialog component per [Figma](https://www.figma.com/file/uUdLhJnS7L7HSoaFw3Pafr/Dialogs?node-id=203%3A596)

📝 Changes from the Figma (cc: @karenliang0):
1. No API `canOutsideClickClose()`: MUI actually recently deprepated their equivalent and is in favor of using `onClose` to handle whether clicking backdrop can close the dialog. So we need the component user to handle if `reason === "backdropClick"` to decide whether to close the dialog or not: https://v4.mui.com/api/dialog/#props
    Screen Shot 2022-01-24 at 4 18 39 PM

2. ~Doesn't seem like we have medium icon, so I'm using "s" for the onClose X button for small Dialog!~ FIXED, thanks to Connor 🎉 !
    Screen Shot 2022-01-24 at 4 16 11 PM

Description:

1. Exposes 5 new components `Dialog`, `DialogTitle`, `DialogPaper`, `DialogContent`, and `DialogActions`, following [MUI Dialog's structure](https://v4.mui.com/components/dialogs/). `DialogPaper` is the component that controls the width of the dialog component via `sdsSize`, its children just take up `DialogPaper`'s width
2. Create `DialogContext` to pass `sdsSize` from top level `<Dialog sdsSize="m" />` to its subcomponents, instead of prop drilling. Currently `DialogTitle` and `DialogPaper` both use `DialogContext` to get `sdsSize`
3. Remove `package-lock.json`, since we use `yarn.lock`
2. Added stories and snapshots:

    1. Screen Shot 2022-01-24 at 3 34 36 PM

    2. demo

    3. Long content:
        demo


clarsen-czi:
Hey @tihuan I noticed you mention we don't have a medium icon button, however we recently added a medium size on the tertiary Icon Button (Storybook here: https://chanzuckerberg.github.io/sci-components/?path=/story/iconbutton--medium-tertiary-icon-button). Is this what you're looking for?

tihuan:
Ahh thanks so much for pointing that out, @clarsen-czi !! Turned out I was using the wrong combos for the IconButtons, and I corrected it after looking at IconButton stories as examples as you said 😆 Thanks again!

mdjmoore (approved):
(◕ᴥ◕)

tihuan:
Thanks for the quick review, @mdjmoore 🙏 !

@karenliang0 PTAL thank you!!

karenliang0:
Everything looks great! The only thing is that when the button positions is left, the Primary Action button should also be on the left instead of the right! Totally my bad because it wasn't stated very clearly in the documentation. Let me know if the below image helps or if you have any questions - thanks so much for working on this Timmy!!

image

tihuan:
Thanks so much for the screenshots and eagle eye, @karenliang0 ! 🦅 👁️ 👍 

Updated storybook accordingly!

Screen Shot 2022-01-27 at 2 00 56 PM

demo

PR #111, opened 2022-01-29
Add button color prop and new error style button

tihuan:
---
title: "type_of_change(scope_of_work): ticket title"
reviewers: "sds-eng, sds-design"
---

## Summary

Copy ticket descirption here

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written
- [ ] Chromatic build verified by @sds-design

### DIFF
--- a/src/core/Button/index.tsx
+++ b/src/core/Button/index.tsx
@@ -9,7 +9,7 @@ import {
   StyledButton,
 } from "./style";
 
-export type ButtonProps = RawButtonProps & ExtraProps;
+export type ButtonProps = Omit<RawButtonProps, "color"> & ExtraProps;
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   (props: ButtonProps, ref): JSX.Element | null => {
--- a/src/core/Button/index.tsx
+++ b/src/core/Button/index.tsx
@@ -1,6 +1,7 @@
 import { ButtonProps as RawButtonProps } from "@material-ui/core";
 import React from "react";
 import {
+  ExtraProps,
   PrimaryMinimalButton,
   RoundedButton,
   SecondaryMinimalButton,
@@ -8,18 +9,11 @@ import {
   StyledButton,
 } from "./style";
 
-interface SdsProps {
-  isAllCaps?: boolean;
-  isRounded?: boolean;
-  sdsStyle?: "minimal" | "rounded" | "square";
-  sdsType?: "primary" | "secondary";
-}
-export type ButtonProps = RawButtonProps & SdsProps;
+export type ButtonProps = RawButtonProps & ExtraProps;
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   (props: ButtonProps, ref): JSX.Element | null => {
-    const sdsStyle = props?.sdsStyle;
-    const sdsType = props?.sdsType;
+    const { color = "primary", sdsStyle, sdsType } = props;
 
     if (!sdsStyle || !sdsType) {
       // eslint-disable-next-line no-console
@@ -44,7 +38,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       case sdsStyle === "rounded" && sdsType === "primary":
         return (
           <RoundedButton
-            color="primary"
+            color={color}
             ref={ref}
             variant="contained"
             {...propsWithDefault}
@@ -53,7 +47,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       case sdsStyle === "rounded" && sdsType === "secondary":
         return (
           <RoundedButton
-            color="primary"
+            color={color}
             ref={ref}
             variant="outlined"
             {...propsWithDefault}
@@ -62,7 +56,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       case sdsStyle === "square" && sdsType === "primary":
         return (
           <SquareButton
-            color="primary"
+            color={color}
             ref={ref}
             variant="contained"
             {...propsWithDefault}
@@ -71,7 +65,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       case sdsStyle === "square" && sdsType === "secondary":
         return (
           <SquareButton
-            color="primary"
+            color={color}
             ref={ref}
             variant="outlined"
             {...propsWithDefault}
@@ -80,7 +74,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       case sdsStyle === "minimal" && sdsType === "primary":
         return (
           <PrimaryMinimalButton
-            color="primary"
+            color={color}
             ref={ref}
             variant="text"
             {...propsWithDefault}
--- a/src/core/Button/style.ts
+++ b/src/core/Button/style.ts
@@ -1,5 +1,5 @@
 import styled from "@emotion/styled";
-import { Button } from "@material-ui/core";
+import { Button, ButtonProps } from "@material-ui/core";
 import {
   fontCapsXxxs,
   getColors,
@@ -13,7 +13,7 @@ export interface ExtraProps {
   isRounded?: boolean;
   sdsStyle?: "minimal" | "rounded" | "square";
   sdsType?: "primary" | "secondary";
-  color?: "success" | "error" | "warning" | "info";
+  color?: "primary" | "secondary" | "success" | "error" | "warning" | "info";
 }
 
 // Please keep this in sync with the props used in `ExtraProps`
@@ -27,12 +27,12 @@ const doNotForwardProps = [
 
 const ButtonBase = styled(Button, {
   shouldForwardProp: (prop) => {
-    return !doNotForwardProps.includes(prop.toString());
+    return !doNotForwardProps.includes(String(prop));
   },
-})<ExtraProps>`
+})`
   box-shadow: none;
-  ${(props) => {
-    const { variant, color: colorProp } = props;
+  ${(props: Props & ExtraProps & Omit<ButtonProps, "color">) => {
+    const { variant, color: colorProp = "primary" } = props;
     const colors = getColors(props);
     const spacings = getSpaces(props);
 
@@ -46,6 +46,10 @@ const ButtonBase = styled(Button, {
     const padding = variant === "outlined" ? outlinedPadding : containedPadding;
 
     const color = colors && colors[colorProp];
+
+    if (!color) {
+      throw new Error(`Button color not found: ${colorProp}`);
+    }
 
     return `
       padding: ${padding};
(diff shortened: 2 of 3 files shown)


src/core/Button/index.tsx
```diff
@@ -1,25 +1,19 @@
...
 import {
+  ExtraProps,
   PrimaryMinimalButton,
   RoundedButton,
   SecondaryMinimalButton,
   SquareButton,
   StyledButton,
 } from "./style";
 
-interface SdsProps {
```
tihuan, line 11: moving this into `style.ts` as `ExtraProps`

src/core/Button/index.tsx
```diff
@@ -1,25 +1,19 @@
...
-  sdsType?: "primary" | "secondary";
-}
-export type ButtonProps = RawButtonProps & SdsProps;
+export type ButtonProps = RawButtonProps & ExtraProps;
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   (props: ButtonProps, ref): JSX.Element | null => {
-    const sdsStyle = props?.sdsStyle;
-    const sdsType = props?.sdsType;
+    const { color = "primary", sdsStyle, sdsType } = props;
```
tihuan, line 16: since MUI Button already accepts `color` prop, we default the `color` selection to `"primary"` since SDS only uses the primary color most of the time

src/core/Button/index.tsx
```diff
@@ -44,7 +38,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       case sdsStyle === "rounded" && sdsType === "primary":
         return (
           <RoundedButton
-            color="primary"
+            color={color}
```
tihuan, line 41: we pass the `color` to each styled button!

src/core/Button/style.ts
```diff
@@ -30,13 +45,15 @@ const ButtonBase = styled(Button, {
 
     const padding = variant === "outlined" ? outlinedPadding : containedPadding;
 
+    const color = colors && colors[colorProp];
```
tihuan, line 48: we pick the color object based on `colorProp`, which defaults to `"primary"` from `index.tsx`

src/core/Button/style.ts
```diff
@@ -30,13 +45,15 @@ const ButtonBase = styled(Button, {
...
+
     return `
       padding: ${padding};
       min-width: 120px;
       height: 34px;
       &:hover, &:focus {
         color: white;
-        background-color: ${colors?.primary[500]};
+        background-color: ${color[500]};
```
tihuan, line 56: once we upgrade to MUI v5, which supports different `color` value other than `primary` and `secondary`, we can remove the custom css in SDS Button that v5 now does for us, and the change will be transparent to SDS library user, because the API is the same 🙆‍♂️

src/core/Button/style.ts
tihuan: also I haven't fixed the type error here 😂

src/core/Button/style.ts
```diff
@@ -8,16 +8,31 @@ import {
...
 const ButtonBase = styled(Button, {
   shouldForwardProp: (prop) => {
-    return !sdsPropNames.includes(prop.toString());
+    return !doNotForwardProps.includes(String(prop));
   },
 })`
   box-shadow: none;
-  ${(props) => {
-    const { variant } = props;
+  ${(props: Props & ExtraProps & Omit<ButtonProps, "color">) => {
```
tihuan, line 34: This fixes the TS issue!

src/core/Button/style.ts
tihuan: Fixed! https://github.com/chanzuckerberg/sci-components/pull/111/files#r794982993

src/core/Button/style.ts
```diff
@@ -30,13 +47,26 @@ const ButtonBase = styled(Button, {
...
+      background-color: ${color[400]};
+      border: ${border[400]};
+      color: white;
       padding: ${padding};
       min-width: 120px;
       height: 34px;
       &:hover, &:focus {
         color: white;
-        background-color: ${colors?.primary[500]};
+        background-color: ${color[600]};
```
jfoo1984, line 69: There is no `500` defined for error color, so I just changed this to use 600 for any color.  The other option would be to add a `500` color for error.  Note that this makes the hover color the same as active below (which was previously `600`).

src/core/Button/style.ts
```diff
@@ -1,23 +1,40 @@
...
+  getBorders,
   getColors,
   getCorners,
   getSpaces,
   Props,
 } from "../styles";
 
-const sdsPropNames = ["isAllCaps", "isRounded", "sdsStyle", "sdsType"];
+export interface ExtraProps {
+  color?: "primary" | "error";
```
jfoo1984, line 13: @tihuan I changed this to only accept the colors that we have used / defined for buttons.  The problem I was running into was that some of the colors aren't defined in our theme at all, or they're missing certain various (`400`, `600`, etc.).  

However, this brings up bit of a philosophical question here.  Are we ok with just defining components that we need as we go (essentially the approach taken here, since I'm mainly working on this to add the error style), which keeps the SDS to the minimal set of components that we actually use?   Or do we want to flesh out all the possible color values for buttons that could be used in MUI so that they will be available in the future if someone needed them (so this would involve having the whole list of color and making sure they work)?

clarsen-czi: Hey Jerry! I want to jump in here—our philosophy with the SDS is for it to be opinionated and specific, meaning we only plan on adding components and their variants as we see the need arise, this is what helps distinguish our design system from simply becoming a sprawling component library. There are a few instances where we have deviated from that (for example anticipating the need for a primary square dropdown button, even though none of our products currently use it, it seemed like there was enough of a potential need to just create it), but overall our approach is to define components that we need as we go. Lmk if you've got any questions!

jfoo1984: Thanks for clarifying @clarsen-czi!  That makes sense and seems like the right approach to me, to just add what we need as we go instead of defining a broad set of components that may never be used.

tihuan: Hey all! Sorry for the late reply 😆 On PTO today!

Yeah I think limiting the color value scope sounds great and matches the actual behavior 👍 

Thanks for updating that!

src/core/Button/style.ts
tihuan: Ahh thanks for pointing this out! 

@clarsen-czi @hthomas-czi Should we define error 500 value in this case? 

Thanks all!

hthomas-czi: I think the ideal behavior would be to use `500` if that value exists, otherwise use `600`.

src/core/Button/style.ts
hthomas-czi: @tihuan Get out of here! You're on PTO! 🤣

src/core/Button/style.ts
clarsen-czi: I agree with Harley—I think we'd ideally use `error500` which doesn't exist. I've just reached out to the designers on the design system team Slack to make sure they're ok with us adding it in. I'll follow up once I hear back

jfoo1984: thanks for chiming in! adding `error500` sounds like a good path.  Then we can consistently have separate hover and active colors for all button types.  I'm happy to add in the `error500` color once we've got it!

jfoo1984: TODO: Add `error500` color (#C61128), and use 500 color for hover.

https://czi-sci.slack.com/archives/C029M3YCFFD/p1643844470244839?thread_ts=1643674010.289539&cid=C029M3YCFFD

src/core/Button/style.ts
```diff
@@ -30,13 +47,26 @@ const ButtonBase = styled(Button, {
...
+      throw new Error(`Button border not found: ${colorProp}`);
+    }
+    if (!color) {
+      throw new Error(`Button color not found: ${colorProp}`);
+    }
+
     return `
+      background-color: ${color[400]};
+      border: ${border[400]};
+      color: white;
```
jfoo1984, line 63: @tihuan, adding these attributes seems to be breaking the secondary button, as that is no longer styled correctly (it ends up getting filled so it just looks like the primary button).  I think we may need to go with just creating a separate override style for the error button.  I don't know if I'll have time to give that a shot, but maybe we can sync up early next week?

src/core/Button/style.ts
```diff
@@ -1,23 +1,40 @@
...
+  sdsType?: "primary" | "secondary";
+}
+
+// Please keep this in sync with the props used in `ExtraProps`
+const doNotForwardProps = [
+  "isAllCaps",
+  "isRounded",
+  "sdsStyle",
+  "sdsType",
+  "color",
```
jfoo1984, line 26: ok, so I think not forwarding `color` to the MUI button is causing issues.  If you look at what was previously on line 11:

```
const sdsPropNames = ["isAllCaps", "isRounded", "sdsStyle", "sdsType"];
```

we did not have color so we did forward it.  I'm beginning to think that the code reuse we're trying to achieve here may not be sustainable in some ways.  I'll start a comment on the overall PR about that.

jfoo1984:
So, I'm having some concerns around this approach where we are trying to generalize the code around colors.  I'm running into some conflicts between the secondary and error styles that can't seem to be resolved.  Basically, to get the error button to display correctly, I need to add this CSS to the base button:
```
      background-color: ${color[400]};
      border: ${border[400]};
      color: white;
```

Those attributes weren't set before.  I believe we were basically relying on passing `color` to MUI, and allowing their defined `primary` and `secondary` to set those attributes.  But when I set those attributes, it breaks the secondary button (it basically turns it into the primary button).

The other thing that breaks the secondary button is that we have to pass through the color prop to MUI as well, because we were relying on styling from that.

Due to these issues, I don't think we can generalize the base button code to support the new error type, and I think it may be easier and clearer to just add error button blocks in the switch statement in `src/core/Button/index.tsx`.

However, that leads to some other issues, and we'd have to check specifically for `color` in addition to `sdsType` and `sdsStyle`, and make sure it's done in the right order so it doesn't fall through to the wrong part of the switch statement.  I'm thinking that we actually should just add `error` as an `sdsType` option, but the switch statement can create it with the needed style

tihuan:
Hey Jerry! Thanks so much for your detailed investigation and writeup 🙏 Updating the button to support more color variants is definitely way more work than expected 😭 

I happened to find a thread talking about the same problem and [their solution](https://github.com/mui-org/material-ui/issues/13875#issuecomment-841869537) seems to work for us too!

I opened a new [draft PR](https://github.com/chanzuckerberg/sci-components/pull/113) to test it out 🙆‍♂️ 

PTAL thank you!

jfoo1984:
Thanks for taking a look and putting the other PR together!  I'll check it out!

tihuan:
Real PR: https://github.com/chanzuckerberg/sci-components/pull/113

PR #115, opened 2022-02-10
refactor(TooltipTable): update Tooltip Component

mdjmoore:
## Summary

**Structural Element (Gene)**
Shortcut ticket: [sh-152186](https://app.shortcut.com/sci-design-system/story/152186/update-tooltiptable-component-for-sds)

**Ticket Description**
- TooltipTable can have only one or multiple sections determined by the needs of the design
- Row labels are left-aligned by default
- Row items are right-aligned by default, though can be left-aligned if needed
- Stories are updated as per [167637](https://app.shortcut.com/sci-design-system/story/167637/update-stories-for-tooltiptablecontent-to-csf-format)

**Additional Changes**
- `alert` prop changed to `contentAlert`. Alert is a reserved term in typescript and was causing errors in the console. Now the extra div doesn't render if no alter is passed, cleaning up our html
- there is no additional spacing around the `contentAlert` subcomponent. Each section has 14px of padding which provides the spacing needed
- new optional prop added `itemAlign`, allows users to right or left align the items in the table. Defaults to right. Will only accept left or right
- `SectionHeader` prop does not exist on component. Instead, section headers are optionally part of the data array that is passed as a prop.
- `data` prop should be an array of objects as follows:
```
data = [
  {
    dataRows: [ {label: "label 1", value: 1}, { label: "Label 2", value: 2}, {label: "Label 3", value: 3}],
    label: "Section Header", // optional
    disabled: false // optional 
  },
  {
    additional sections as needed
  },
];
```
- new optional prop added to Tooltip, `followCursor`, defaults to false
- Table story removed from Tooltip section in storybook

## Checklist

- [x] Test Story in Storybook
- [x] Tests written
- [ ] Chromatic build verified by @sds-design


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,3 @@
----
-title: "type_of_change(scope_of_work): ticket title"
-reviewers: "sds-eng, sds-design"
----
-
```
mdjmoore, line 5: Github issues supports frontmatter, pull requests do not, so remove it

src/core/Tooltip/index.tsx
```diff
@@ -77,13 +78,42 @@ export default forwardRef(function Tooltip(
...
   return (
     <RawTooltip
       classes={{ arrow, tooltip }}
       leaveDelay={leaveDelay}
       interactive
       title={content}
       ref={ref}
+      onMouseMove={
+        followCursor
+          ? (e) => setPosition({ x: e.pageX, y: e.pageY })
```
tihuan, line 92: We probably just need to keep an eye on `onMouseMove()`, since it is called without throttling, so we're setting state as fast as every 6ms. 

For perf, we could throttle the set state to be roughly every 16ms to match 60fps, but that would require adding `lodash/throttle` to SDS as a dependency, I'm not sure it's worth it right now given the tooltip rendering seems cheap

But in the future, if products do see perf issue with this, we can check if throttling will help!

tihuan (approved):
Looks super awesome!! Thanks so much for adding the feature and tests, @mdjmoore 🚀 🌟 !!

PR #123, opened 2022-02-22
chore(storybook): upgrade storybook

tihuan:
Upgrade storybook! PTAL thank you 😄 🙏


mdjmoore (approved):
( ◑ ᗜ ◑ )

jfoo1984:
late comment on this, but do we want to consider pinning the version to a specific version and not allow patch version updates with `^`?  My reasoning on this is that this will insure consistency in whatever environment someone is installing the repo, and forces us to be more intentional about updating our package versions.

tihuan:
Hey @jfoo1984 ! Ooh that's a good question 💡 Patch versions don't have breaking changes, so it feels like it's pretty low risk to have slightly different patch versions across individuals' local environments? Or are there other things we should consider? Thank you!

jfoo1984:
That's a good point!  The main considerations I'd have here are absolute consistency across environments, and potential thrash of the `package-lock.json` (I'm not sure if that file is updated when higher patch versions are allowed).

PR #132, opened 2022-03-02
fix(types): update type files to avoid name collisions

tihuan:
Our type definition file had `Props` global name collisions, which cause TS to infer the wrong types when type checking. This PR fixes the most common cause of this.

⚠️ **BREAKING CHANGE: `import { Props } from 'czifui'` is now `import { CommonThemeProps } from 'czifui'`**

The following is an example that the type checking works as expected again!

Screen Shot 2022-03-01 at 5 50 13 PM

Discussion:
The bigger issue is that **sometimes** `rollup-plugin-ts` will infer the types incorrectly, but most of the time it will know to add incremental suffix $0, $1, $2, etc. for interfaces with the same name. It's just that in the case of `Props`, it didn't do so for the most critical case. I haven't had time to figure out why yet!

## Summary

Copy ticket descirption here

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


mdjmoore (approved):

mmmmmmaya:
I know I'm late to the party, but this looks great. Thanks for following up so quickly!

PR #146, opened 2022-03-16
feat(accordion): add component and sub components

mdjmoore:
## Summary

**Structural Element (DNA)**
Shortcut ticket: [sh-184852](https://app.shortcut.com/sci-design-system/story/184852/implement-accordion)

Props:
- `togglePosition`: "left" | "right" default
- `expanded`: true | false
- `useDivider`: true | false

New Subcomponents:
- `AccordionHeader`: has subtitle prop
- `AccordionDetails`: wrapper to hold any elements within

## Checklist

- [x] Test Story in Storybook
- [x] Tests written


src/core/Accordion/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+      setExpanded(isExpanded ? panel : false);
+    };
+  return (
+    <StyledAccordion
+      id={id}
+      square
+      useDivider={useDivider}
+      togglePosition={togglePosition}
+      expanded={expanded === id}
+      onChange={handleChange(id)}
```
mmmmmmaya, line 23: should this be `() => handleChange(id)`? Otherwise it only executes once, right?

mdjmoore: onChange is already a callback function, so we're passing the function that it's going to call. If this was just a regular react component, you're correct, but MUI accounts for the callback

mmmmmmaya (approved):

mmmmmmaya (approved):

PR #157, opened 2022-04-13
fix(button): v5 upgrade, fix minimal secondary styles, add focus-visible

mdjmoore:


src/core/Button/index.tsx
```diff
@@ -96,7 +96,7 @@ const Button = React.forwardRef(
       case sdsStyle === "minimal" && sdsType === "secondary":
         return (
           <SecondaryMinimalButton
-            color="default"
+            color="primary"
```
tihuan, line 99: ah I think we might need to implement the "default" style or something to match https://chanzuckerberg.github.io/sci-components/?path=/story/button--minimal-secondary 🙏 Thank you!

PR #160, opened 2022-04-13
fix(dialog): fix CSS specificity for v5, remove a11y errors, snaps

mdjmoore:


src/core/DialogPaper/style.ts
```diff
@@ -13,20 +13,22 @@ const doNotForwardProps = ["sdsSize"];
...
-    const spaces = getSpaces(props);
-    const corners = getCorners(props);
-
-    return `
-      max-height: calc(100vh - ${2 * (spaces?.xxl || 0)}px);
-      border-radius: ${corners?.m || 0}px;
-      padding: ${spaces?.xxl || 0}px;
-    `;
-  }}
+  & {
```
mdjmoore, line 16: wrapper needed for any styles that are getting lost in MUI default styles

src/core/DialogTitle/index.tsx
```diff
@@ -22,8 +22,8 @@ const DialogTitle = forwardRef(function DialogTitle(
       {children || (
         <>
           {onClose && <CloseButton onClick={onClose} />}
-          <Title variant="h6">{title}</Title>
```
mdjmoore, line 25: because the v5 dialog title renders as an h2, nesting an h6 inside of it generates a11y errors. Removed h6 in favour of p tags

tihuan (approved):
LGTM! Thanks so much, @mdjmoore 🎉 !!

PR #169, opened 2022-04-14
fix(complexfilter): Update ComplexFilter

tihuan:
## Summary

demo

This change touches `ComplexFilter`, `MenuSelect`, `InputSearch`, and `MenuItem`, since `ComplexFilter` uses all of them!

TODO: I haven't had a chance to fix `MenuSelect` and `Dropdown` yet, but they are both super similar to `ComplexFilter`, so the same fix should work there too!

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/src/core/ComplexFilter/style.ts
+++ b/src/core/ComplexFilter/style.ts
@@ -1,4 +1,5 @@
 import { Paper } from "@mui/material";
+import { autocompleteClasses } from "@mui/material/Autocomplete";
 import Popper from "@mui/material/Popper";
 import { styled } from "@mui/material/styles";
 import {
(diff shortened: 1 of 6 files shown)


src/core/ComplexFilter/style.ts
```diff
@@ -14,57 +15,56 @@ export const Wrapper = styled("div")`
...
+  & {
+    ${fontHeaderXs}
 
-  .MuiAutocomplete-popperDisablePortal {
-    position: relative;
-    width: 100% !important;
-  }
+    .${autocompleteClasses.popperDisablePortal} {
+      position: relative !important;
+      transform: none !important;
```
mmmmmmaya, line 23: is `important` necessary here, or can we get away with `&&`?

mmmmmmaya (approved):

src/core/ComplexFilter/style.ts
tihuan: Thanks so much for the quick review, @mmmmmmaya 🙏 !

Just tested it out! I think since `Popper` applies the `transform` as inline style, it seems that `!important` is the only way to fight it 😭 I just added a comment to document this too!

tihuan: @bethbertozzi for fixing Autocomplete dropdown misalignment, `transform: none !important;` was the key! I think Dropdown and MenuSelect could probably reuse most if not all fixes in this PR 😆 

Thank you!

bethbertozzi: thanks @tihuan!

PR #171, opened 2022-04-15
feat(inputtoggle): add an inputtoggle

mmmmmmaya:
### Summary
- **What:** Add `InputToggle` component
- **Ticket:** [[sc-157940]](https://app.shortcut.com/sci-design-system/story/157940)
- **Design:** https://www.figma.com/proto/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=1876%3A21139&scaling=min-zoom&page-id=642%3A1

### Demos
after

### Checklist
- [x] I merged latest `main`
- [x] I added labels to my PR
- [x] I either asked design for a review or hid my changes behind a feature flag
- [x] I added relevant unit tests
- [x] Test Story in Storybook

### DIFF
--- a/src/core/InputToggle/style.ts
+++ b/src/core/InputToggle/style.ts
@@ -1,0 +1,154 @@
+import styled from "@emotion/styled";
+import { Switch, SwitchProps } from "@material-ui/core";
+import {
+  CommonThemeProps,
+  fontBodyXs,
+  getBorders,
+  getColors,
+  getCorners,
+  getSpaces,
+} from "../styles";
+
+export interface ExtraProps extends SwitchProps, CommonThemeProps {
+  offLabel?: string;
+  onChange?(e: React.ChangeEvent): void;
+  onLabel?: string;
+}
+
+const toggle = (props: ExtraProps) => {
+  const { disabled } = props;
+  const corners = getCorners(props);
+  const spaces = getSpaces(props);
+  const TOGGLE_HEIGHT = 18;
+
+  return `
+    cursor: ${disabled ? "default" : "pointer"};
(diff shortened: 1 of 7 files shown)


src/core/InputToggle/style.ts
```diff
@@ -0,0 +1,154 @@
...
+  const corners = getCorners(props);
+  const spaces = getSpaces(props);
+  const TOGGLE_HEIGHT = 18;
+
+  return `
+    cursor: ${disabled ? "default" : "pointer"};
+    border-radius: ${corners?.l}px;
+    height: 26px;
+    width: 62px;
+    line-height: 18px;
```
j-x-han, line 29: @chanzuckerberg/sds-design Are these values we'll reuse/should we define these as bases somewhere?

j-x-han (approved):
LGTM overall, just a question for design about some of the hardcoded values

src/core/InputToggle/style.ts
Janeece: Hmm I can imagine there are cases where we'd want the toggle to be able to increase the width if needed (like if we needed to use a longer word). We haven't really discussed that or how to handle hardcoded cases. 

Would we be able to allow products using this component to override the dimensions if needed?

Janeece (approved):
@mmmmmmaya I think this looks great! Not sure what the outcome will be from Julie's Question.

src/core/InputToggle/style.ts
mmmmmmaya: I believe comsumer could override with width, as written, by importing the sds toggle and then using emotion/styled.

PR #183, opened 2022-05-10
feat(callout): allow consumer to conditionally override callout icon

mmmmmmaya:
### Summary
- **What:** Allow a consumer to conditionally assign a custom icon to callout
- **Why:** I have a component that is based on Callout, which wants to provide a custom icon _sometimes_. As is, if the component passes `icon={iconComponent}`, it works, but then if `iconComponent` is undefined, sds Callout reverts to using the mui default icon because `icon={undefined}` gets spread in sds Callout via `{...rest}`. With this change, if my component passes `icon={undefined}` (where undefined is determined programmatically), we still use the icon as dictated by sds style guidelines, rather than mui
- **Ticket:** [[sc-180631]](https://app.shortcut.com/genepi/story/180631)

#### Before: 
Screen Shot 2022-05-09 at 5 17 13 PM

#### After: 
Screen Shot 2022-05-09 at 5 17 38 PM

### Checklist
- [x] I merged latest `main`
- [x] I added labels to my PR
- [ ] I either asked design for a review or hid my changes behind a feature flag

### DIFF
--- a/src/core/Callout/index.tsx
+++ b/src/core/Callout/index.tsx
@@ -8,6 +8,7 @@ import { StyledCallout } from "./style";
 export interface CalloutProps {
   autoDismiss?: boolean | number;
   dismissed?: boolean;
+  icon?: React.ReactNode;
   intent: "info" | "error" | "success" | "warning";
 }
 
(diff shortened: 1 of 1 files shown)


tihuan (approved):
Ahh great insight! Thanks for the explanation, Maya! LGTM 🚀 !!

PR #184, opened 2022-05-10
fix(tooltip): fix styling broken from migration to mui5

bethbertozzi:
## Summary
Merge fixes on broken components, not including dropdowns. Fixed input dropdown, menu, notification, and tooltip. Masoud is refactoring MenuSelect on main, so I am waiting until that is merged into main, and then beta, to migrate those related components. 

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/src/core/InputDropdown/style.ts
+++ b/src/core/InputDropdown/style.ts
@@ -1,4 +1,6 @@
 import { css, SerializedStyles } from "@emotion/react";
+import { autocompleteClasses } from "@mui/material/Autocomplete";
+import Popper from "@mui/material/Popper";
 import { styled } from "@mui/material/styles";
 import Button from "../Button";
 import {
@@ -274,4 +276,13 @@ export const StyledCounter = styled("span")`
     padding: 1px ${spacings?.xs}px;
   `;
   }}
+`;
+
+export const StyledPopper = styled(Popper)`
+  & {
+    .${autocompleteClasses.popperDisablePortal} {
+      position: relative !important;
+      transform: translate3d(5px, 0px, 0px) !important;
+    }
+  }
 `;
--- a/src/core/InputDropdown/style.ts
+++ b/src/core/InputDropdown/style.ts
@@ -126,6 +126,10 @@ const square = (props: InputDropdownProps): SerializedStyles => {
     border-radius: ${corners?.m}px;
     height: 34px;
     min-width: 90px;
+
+    > span:first-of-type {
+      margin-right: 0px;
+    }
   `;
 };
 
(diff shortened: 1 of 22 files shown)


clarsen-czi (changes requested):
I added comments on some of the builds in the chromatic review, but there were some overall themes I noticed, so decided to put them here instead of commenting on every build.

For all Buttons can you ensure that after the Button has been pressed and the user moves their cursor off of the Button it returns to its default state (currently the hover state stays in place causing the Button to be filled in).

For all DropdownMenus that open, can you ensure that the left edge of the menu is left-aligned with the element triggering it (whether that's an InputDropdown, ButtonIcon, etc.)

src/core/Tooltip/style.ts
```diff
@@ -91,13 +92,14 @@ export const tooltipCss = (props: ExtraProps): string => {
...
 export const arrowCss = (props: ExtraProps): string => {
-  const { inverted, sdsStyle } = props;
+  const { inverted, sdsStyle, arrowOffset } = props;
 
   const borders = getBorders(props);
 
   return css`
     &.MuiTooltip-arrow {
       color: ${inverted || sdsStyle === "dark" ? "black" : "white"};
+      left: ${arrowOffset}px !important;
```
mmmmmmaya, line 102: can we create a more specific selector rather than using `!important`?

src/core/InputDropdown/style.ts
```diff
@@ -271,3 +282,12 @@ export const StyledCounter = styled("span")`
   `;
   }}
 `;
+
+export const StyledPopper = styled(Popper)`
+  & {
+    .${autocompleteClasses.popperDisablePortal} {
+      position: relative !important;
```
mmmmmmaya, line 289: same here re: important

mmmmmmaya:
looks good besides maybe choose some higher specificity selectors to avoid use of important

src/core/Tooltip/style.ts
bethbertozzi: @mmmmmmaya correct me if I'm wrong but I think I need it here to overwrite the `left` styles applied at rendering, happy to look into it more if not 
Screen Shot 2022-05-19 at 2 20 55 PM

src/core/InputDropdown/style.ts
bethbertozzi: This popper is just used in the story and not a part of the component library and won't be used by consumers so I thought it would be ok, I can still look into it if not

clarsen-czi (changes requested):
Looking good, Beth! A few more changes needed.

bethbertozzi:
@clarsen-czi I am not able to respond to your comments in chromatic :/ The accordion is not yet migrated and has a ticket to do so: https://app.shortcut.com/sci-design-system/story/199604/migrate-accordion. I investigated the tooltip issue and found that it is only persisting on chrome, it requires more investigation and is not a result of the migration (it appears to be happening on main: https://chanzuckerberg.github.io/sci-components/?path=/story/inputsearch--default), so I made a new ticket for it here: https://app.shortcut.com/sci-design-system/story/200241/remove-extra-tooltip-from-inputsearch-and-inputtext

PR #189, opened 2022-05-19
refactor(all): Masoudamjadi/merge main into beta

masoudmanson:
## Summary

Updating `beta` branch with the changes from `main`


bethbertozzi (commented):
i think the base branch should be beta, right?

masoudmanson:
> i think the base branch should be beta, right?

Ooh, you're right. It should be based on the `beta`. This is weird.

PR #191, opened 2022-05-26
fix(inputsearch): remove extra data-test-id from index file

masoudmanson:
Removes Warning: React does not recognize the `data-testId` prop on a DOM element.


tihuan (approved):
Thanks for catching this!

PR #192, opened 2022-05-27
style(MenuItem): removes extra min-height css property from menuItem component

masoudmanson:
## Summary

**MenuItem**
removes extra min-height css property from menuItem component


tihuan (approved):
LGTM! Thanks, Masoud 😄 🙏 !

PR #195, opened 2022-06-01
refactor(tag): updates the hover and active logic for custom colored tags 

masoudmanson:
## Summary

**Tag**
Using MUI's `darken()` function to replace the `::after` pseudo-element previously used to change the `:hover` and `:active` background-colors for custom colored tags.


tihuan (approved):
Ooh nice refactor! Do we need design's eyes on the change, just in case the darkened colors need adjustment? (It looks good to me, but just in case lol)

Thank you!

masoudmanson:
> Ooh nice refactor! Do we need design's eyes on the change, just in case the darkened colors need adjustment? (It looks good to me, but just in case lol)
> 
> Thank you!

Sure, why not 🤓

clarsen-czi (approved):
Thanks Masoud!

PR #196, opened 2022-06-01
feat(menuselect, dropdownmenu): Add option to prevent Dropdown from clearing search bar on select

masoudmanson:
## Summary

**MenuSelect, DropDownMenu**
Shortcut ticket: [sh-193551](https://app.shortcut.com/sci-design-system/story/193551/add-option-to-prevent-dropdown-from-clearing-search-bar-on-option-select)

"The search bar -- when search is true -- clears on selection of any item in the Dropdown. This isn't exactly a bug, but for the component we're working on in CZ GEN EPI we really need a way to change the behavior so that the search bar keeps its entered search term and the filtered options remain the same before and after selection of an item."

+ [Added] KeepSearchOnSelect option for menuSelect
+ [Added] KeepSearchOnSelect option for dropDownMenu
+ [Changed] Minor styling updates for MenuSelect component

If you set `keepSearchOnSelect` as `false`, after clicking on an option, the search box would be cleared too. This is how it was previously.

KeepSearchOnSelect = False

But setting `keepSearchOnSelect` as `true` would prevent clearing the search bar on option select.

KeepSearchOnSelect = true

### DIFF
--- a/src/core/DropdownMenu/index.stories.tsx
+++ b/src/core/DropdownMenu/index.stories.tsx
@@ -202,6 +202,9 @@ export default {
     hasSections: {
       control: { type: "boolean" },
     },
+    keepSearchOnSelect: {
+      control: { type: "boolean" },
+    },
     multiple: {
       control: { type: "boolean" },
     },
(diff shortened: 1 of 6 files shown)


tihuan (approved):

clarsen-czi (approved):
LGTM!! Thanks Masoud!

PR #199, opened 2022-06-03
refactor(tag): changed FilterTag to TagFilter and style changes

masoudmanson:
## Summary

**Tag, TagFilter**

- Change the FilterTag component's name to TagFilter. 
- Some styling changes based on Connor's previous reviews.

## Checklist

- [x] Test Story in Storybook
- [x] Tests written


src/core/Tooltip/index.stories.tsx
```diff
@@ -75,6 +75,8 @@ Dark.parameters = {
 
 Dark.args = {
   arrow: true,
+  // delay allows for font to load and prevents chromatic from constantly creating new baselines
```
mmmmmmaya, line 78: Is there a way we can set this as a config option in chromatic, rather than setting it for several components?

clarsen-czi (changes requested):
Everything looks great Masoud! Only comment is to make sure there is no space in the name "TagFilter" (currently you have it as "Tag Filter")

src/core/Tooltip/index.stories.tsx
masoudmanson: Thanks for the review @mmmmmmaya 
Yes we can, I'll update the PR 🤓

clarsen-czi (approved):
Looks great! Thanks Masoud!!

PR #202, opened 2022-06-06
chore(package.json): update ip package

jfoo1984:
`yarn start` was erroring out in the `ip` package, as shown below.  If this is a broader issue that others are experiencing, forcing `yarn` to use a newer `ip` package via the `resolutions` config seems to solve the issue.

```
❯ yarn && yarn start
yarn install v1.22.17
[1/4] 🔍  Resolving packages...
success Already up-to-date.
$ husky install
husky - Git hooks installed
✨  Done in 0.62s.
yarn run v1.22.17
$ start-storybook -p 6006
info @storybook/react v6.4.22
info
info => Loading presets
Warning: React.createFactory() is deprecated and will be removed in a future major release. Consider using JSX or use React.createElement() directly instead.
ERR! TypeError: details.family.toLowerCase is not a function
ERR!     at /Users/jfu/sci-components/node_modules/ip/lib/ip.js:385:39
ERR!     at Array.filter (<anonymous>)
```

## Summary

Copy ticket descirption here

## Checklist

- [x] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,9 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1"
   },
+  "resolutions": {
+    "ip": "1.1.6"
+  },
   "scripts": {
     "build": "yarn sd-build && rollup -c",
     "build-watch": "rollup -c -w",
--- a/yarn.lock
+++ b/yarn.lock
@@ -9016,10 +9016,10 @@ ip-regex@^4.1.0:
   resolved "https://registry.yarnpkg.com/ip-regex/-/ip-regex-4.3.0.tgz#687275ab0f57fa76978ff8f4dddc8a23d5990db5"
   integrity sha512-B9ZWJxHHOHUhUjCPrMpLD4xEq35bUTClHM1S6CBU5ixQnkZmwipwgc96vAd7AAGM9TGHvJR+Uss+/Ak6UphK+Q==
 
-ip@^1.1.5:
-  version "1.1.5"
-  resolved "https://registry.npmjs.org/ip/-/ip-1.1.5.tgz"
-  integrity sha1-vd7XARQpCCjAoDnnLvJfWq7ENUo=
+ip@1.1.6, ip@^1.1.5:
+  version "1.1.6"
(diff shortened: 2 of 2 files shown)


tihuan:
Thanks so much for the PR, @jfoo1984 🤩 🙏

Do we know which library depends on `ip`? Assuming our library doesn't have a direct dependency on it. And if so, can we check if it's possible to upgrade that library, so we update both the direct dependency and the indirect dependency (`ip`)?

Also can other people reproduce this? Given that we all have the same `yarn.lock` from the repo, I wonder if everyone is getting the same error or is it limited to certain machines like the ones using M1 or something?

Thank you!

jfoo1984:
@tihuan this is what I see using the ip package

```
❯ npm ls ip
czifui@0.0.0-development /Users/jfu/sci-components
├─┬ @storybook/react@6.4.22
│ └─┬ @storybook/core@6.4.22
│   └─┬ @storybook/core-server@6.4.22
│     └── ip@1.1.6 deduped
├─┬ semantic-release@18.0.0
│ └─┬ @semantic-release/npm@8.0.0
│   └─┬ npm@7.24.1
│     └─┬ make-fetch-happen@9.1.0
│       └─┬ socks-proxy-agent@6.1.0
│         └─┬ socks@2.6.1
│           └── ip@1.1.5
└─┬ socks@2.6.1 extraneous
  └── ip@1.1.6
```

I agree it would definitely be good to hear if anyone else is running into this as well.  In case it's relevant, I'm running node version `18.2.0`.  Could that possible be an issue?

jfoo1984:
node version That could be the problem, I'm on node 18.2.0.  Is node v14.17.4 the recommended version for running the project?

jfoo1984:
I switched to node 16.15.1, which I believe is the current stable LTS version, and that seems to work.  

Would it be worth adding a .node-version file to indicate the node version?  That's the file that the node version manager `nodenv` uses to autoswitch the node version, but we could add other files as well if other version managers use a different file.

mmmmmmaya:
I'm in favor of the .node-version file. thanks for the suggestion!

Issue #88, opened 2021-10-29
Request: way to reduce/silence deprecation warnings


vincent-czi:
The deprecation warnings on components/functions are definitely important, but it's possible to have them fire off to a huge degree. Currently, when I load the Aspen app, I get ~300 deprecation warnings all at once. And any subsequent page interaction causes another 50-200 to show in the console log. It makes any logging I need to do for dev work extremely hard to find amongst all the warnings.

In this case, the two deprecation warnings that are blowing up my console are `Buttons without sdsStyle or sdsType props will be deprecated` and `getSpacings() is deprecated and may be removed in a future release`. Since they're both used a ton throughout the app, there's a huge amount of warnings popping off with every render event.

We certainly do need to address our deprecation warnings -- don't get me wrong! -- but pretty frequently we need to postpone handling that for a couple weeks while we're in the middle of something else. I want some way to quiet those warnings when I need to postpone those sort of fixes.

Two possible approaches I can think of right now:
- Some kind of flag that completely disables deprecation warnings from czifui. Possibly have a single console warning that says once on load, "hey! you have disabled deprecation warnings, that's dangerous, be careful".
- Have the deprecation warnings only fire once for each kind of thing. So for us, we'd see a single `Buttons` deprecation warning and a single `getSpacings`, but that's it.

In either case with the two above, the big thing is that I wouldn't be seeing them show up every single render, potentially burying log events I care about noticing.

Thank you!

vincent-czi:
Okay, I just realized I can also type `-deprecated` in the console log filter to strip all the warnings while I'm working, so while I think the above has some merit, it's probably not super important to prioritize soon.

mdjmoore:
`getSpacings()` is addressed as part of this commit: https://github.com/chanzuckerberg/sci-components/commit/6fd0dd28f7c8b790e7f74906f2ce32a1c2179a04 which hasn't been released yet. We can look into adding a bug fix to address button deprecation warnings and have it as hopefully part of a near future relase

mdjmoore:
ticket created [here](https://app.shortcut.com/sci-design-system/story/171875/request-way-to-reduce-silence-deprecation-warnings-88)

jfoo1984:
It looks like there was some logic added in #155 that reduces / limits the number of warnings.  I did a bit of testing by adding multiple buttons that triggered the deprecation warning on a page, and only got one warning.

[...]

PR #203, opened 2022-06-06
refactor(button,chip,menuselect): Consolidate one-time warnings and remove deprecated code in docs

jfoo1984:
## Summary

- Consolidate deprecation warnings / and logic
- Update storybook examples that use deprecated `Button` syntax
Shortcut ticket: [sh-171875](https://app.shortcut.com/sci-design-system/story/171875/request-way-to-reduce-silence-deprecation-warnings-88)

>From Vincent: https://github.com/chanzuckerberg/sci-components/issues/88
>
>`getSpacings()` is dealt with but hasn't been released yet (oct 29th). We should find a way to only show warnings once on initial load and not every time a component with a warning is rendered
>
>ideas:
>- add a warning ID and only show new IDs when rendering 
>- have a Set of warnings and only print that set to the console on render

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


mmmmmmaya (approved):

jfoo1984:
@mmmmmmaya I made some substantive changes, because we expanded the scope of the story to include consolidating the deprecation warnings.

Do we have a pattern / requirement for testing code outside components?  I started working on tests for the new code in `warnings.ts`

src/common/warnings.ts
```diff
@@ -0,0 +1,35 @@
...
+      "Warning: MenuSelect will be deprecated and replaced with <DropdownMenu />",
+  },
+};
+
+export const showWarningIfFirstOccurence = (warningType: SDSWarningTypes) => {
+  if (
+    !(<string>warningType in SDS_WARNINGS) ||
+    SDS_WARNINGS[warningType].hasWarned
+  )
+    return;
```
jfoo1984, line 29: Because of typescript, you can't actually call this method if the warning type isn't in `SDSWarningTypes`, but I think it's potentially possible that we add a warning type without creating the field and object in the `SDS_WARNINGS` object.

mmmmmmaya: I think it would be better to have it be more restrictive. What do you think? For example, it would guide developers towards adding the warning here, so other developers can see which types of warnings exist at a glance. (or maybe I'm missing the problem 😓)

mmmmmmaya:
I don't believe we have an existing convention for that, but I think it would be ok to include a test file in the same directory as the file you are testing. That's how we do it for components, and it feels good to me to stay consistent in that pattern with non-component cases as well. So, a `warning.test.ts` file in `src/common`.

src/common/warnings.ts
mmmmmmaya: Besides this one comment, this PR lgtm.

jfoo1984: So I guess there isn't really a problem per se, but my comment was more to explain why this check is helpful.  But now that you mention it, it might be worth adding a warning that the warning type isn't defined, and adding that to the object to prevent it from showing over and over.

.node-version
```diff
@@ -0,0 +1 @@
+16.15.1
```
jfoo1984, line 1: current active LTS version of node.  with node 18 installed, I ran into dependency installation issues

src/common/warnings.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  it("does not show warning on subsequent calls", () => {
+    showWarningIfFirstOccurence(SDSWarningTypes.ChipDeprecated);
+    showWarningIfFirstOccurence(SDSWarningTypes.ChipDeprecated);
+    showWarningIfFirstOccurence(SDSWarningTypes.ChipDeprecated);
+
+    expect(consoleWarnSpy).toHaveBeenCalledTimes(1);
+  });
+});
```
jfoo1984, line 28: @mmmmmmaya added a spec for the new warnings file!

src/common/warnings.ts
```diff
@@ -0,0 +1,41 @@
...
+  },
+  [SDSWarningTypes.MenuSelectDeprecated]: {
+    hasWarned: false,
+    message:
+      "Warning: MenuSelect will be deprecated and replaced with <DropdownMenu />",
+  },
+};
+
+export const showWarningIfFirstOccurence = (warningType: SDSWarningTypes) => {
+  if (!(<string>warningType in SDS_WARNINGS)) {
```
jfoo1984, line 25: FYI, if you actually add a new type in the `SDSWarningTypes` enum without adding it to the `SDS_WARNINGS` object, there are a number of typescript errors that are raised in this method, but this warning will make it explicit in the browser as well.

PR #205, opened 2022-06-16
feat: SDS Tailwind Style Dictionary Formatter

codemonkey800:
## Summary

Shortcut ticket: [sh-196360](https://app.shortcut.com/sci-design-system/story/196360/tailwind-config-using-sds-design-tokens)

Adds a new `style-dictionary` formatter for generating a [Tailwind configuration](https://tailwindcss.com/docs/configuration) that can be imported into a Tailwind based application so that SDS styles can be used. Since this generates a Tailwind compliant configuration, any editor plugins will automatically complete classes.

### Setup

Users will need to import the SDS Tailwind config into their configuration:

```js
// tailwind.config.js
const sds = require("czifui/dist/tailwind.json");
module.exports = {
  mode: "jit",
  content: ["./src/**/*.{tsx,scss}"],
  theme: {
    extend: {
      ...sds,
      width: {
        ...sds.width,
        "app-modal-width": "500px",
      },
      height: {
        ...sds.width,
        "app-modal-height": "200px",
      },
    },
  },
};
```

### Colors

image

image

### Spacing

image

### Corners

image

### Letter Spacing

image

### Shadows

image

### Borders

Borders in Tailwind are broken down into smaller design tokens, so it'll take more classes to get an equivalent border in Tailwind.

For example, the border token `$sds-borders-error-400` has the value `1px solid #dc132c`. To recreate this in Tailwind:

```tsx
<div className="border border-solid border-sds-error-400" />
```

### Fonts

Similar to borders, font tokens are broken down into smaller tokens for Tailwind. For example, the token `$sds-font-body-xxs-600-font` has the value `600 12px/18px 'Open Sans', 'OpenSans-SemiBold'`. To recreate this in Tailwind:

```tsx
<div className="text-sds-body-xxs leading-sds-body-xxs font-sds-semibold" />
```


README.md
```diff
@@ -114,6 +114,65 @@ export const Tag = styled.div`
 }
 ```
 
+5. Tailwind - Tailwind compliant configuration for the `defaultTheme` to use if your app uses TailwindA
```
mmmmmmaya, line 117:
```suggestion
5. Tailwind - Tailwind compliant configuration for the `defaultTheme` to use if your app uses Tailwind
```

src/common/styles-dictionary/tailwindFormat.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+      }
+
+      return result;
+    }
+
+    function transformColor(tokens) {
+      const colors = {};
+
+      for (const [name, colorTokens] of Object.entries(tokens)) {
```
mmmmmmaya, line 35: what is the difference between colorToken and token? is it possible to rename them to something more english-y?

mmmmmmaya (approved):
I think this basically looks ok to me. Wondering if there needs to be any communication to the team about how to keep this up to date or if tailwind will automatically stay in line with the global styles we have defined.

codemonkey800:
Going to skip making this a plugin for now, both this and the plugin approach provide the same thing, except this approach doesn't require us to install taiwindcss to use their plugin interface. We can look into making it a plugin again if we encounter a use-case for it

mmmmmmaya:
sgtm 🚀

PR #209, opened 2022-06-27
refactor(button): change to csf format

bethbertozzi:
## Summary
Button does not have 'error' color built, created ticket here: https://app.shortcut.com/sci-design-system/story/204195/add-error-color-to-button

Shortcut ticket: [sh-XXXX](https://app.shortcut.com/sci-design-system/story/167622/update-stories-for-button-to-csf-format)

In addition, there should be 3 stories per component: Default (all props are interactable via storybook controls, will occasionally include some external components to trigger effects and states), Live Preview (a number of components as defined by our ZeroHeight implementation—see below for which components to include in each LivePreview view), and Test (a single component).

Default and Live Preview should have the following parameters set so a snapshot is only taken of the single component:

Default.parameters = {
  snapshot: {
    skip: true,
  },
};

LivePreview.parameters = {
  snapshot: {
    skip: true,
  },
};
 Wrap
Most props should auto populate

Live Preview

There should be an individual Live Preview story for each Button variant since these will be imported into Zeroheight within each variant section. See attached screenshots for visual guidance of how the Live Preview should be laid out:

Rounded Button
Primary, Primary + Icon, Secondary, Secondary + Icon
Spacing as follows: 10px between each variant
Rounded Button.png

Square Button
Primary, Primary + Icon, Secondary, Secondary + Icon
Spacing as follows: 10px between each variant
Square Button.png

Text Button
Primary, Primary + Icon, Secondary
Spacing as follows: 24px between each variant
Text Button.png

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


clarsen-czi (changes requested):
Looking great @bethbertozzi ! Just a few changes to be made and then we should be good.

clarsen-czi (changes requested):
Unfortunately it looks like the left/right padding is still on the minimal button. Once we get that figured out this one will be good to go!

bethbertozzi:
@clarsen-czi I tried just displaying a minimal button with no grid added in the live preview, it looks like the button itself has padding. Is this what you are referring to? I can see how big of a job it is to go through and change the styling, but it is probably cleaner to make another ticket for it so the change is tracked and so it doesn't hold up the CSF formatting changes
[Screen Shot 2022-07-08 at 11.44.26 AM.pdf](https://github.com/chanzuckerberg/sci-components/files/9073370/Screen.Shot.2022-07-08.at.11.44.26.AM.pdf)

clarsen-czi (approved):
Thanks @bethbertozzi ! I agree that we should have a new ticket for removing the padding on the button. Can you create that ticket? Thanks!

PR #215, opened 2022-07-06
refactor(iconbutton): change to csf

bethbertozzi:
## Summary

Shortcut ticket: [sh-XXXX](https://app.shortcut.com/sci-design-system/story/167627/update-stories-for-buttonicon-to-csf-format)

Default.parameters = {
  snapshot: {
    skip: true,
  },
};

LivePreview.parameters = {
  snapshot: {
    skip: true,
  },
};
 Wrap
Most props should auto populate

Live preview as shown here with 24px between each variant: https://www.figma.com/file/LS5AaNnOmjJ8ZDIuq0Tl9E/Zeroheight-Assets?node-id=21%3A60013

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


src/core/IconButton/index.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import { Args, Story } from "@storybook/react";
 import React from "react";
 import Icon from "../Icon";
+import { getSpaces } from "../styles";
```
masoudmanson, line 4: `getSpaces` needs `theme` as a parameter and since we don't have theme object in storybook stories, I think we can directly import `defaultTheme` object and use it to get `spaces`.

```
import { defaultAppTheme } from "../styles";
```

src/core/IconButton/index.stories.tsx
```diff
@@ -28,72 +29,158 @@ export default {
...
-  disabled: false,
-  icon: <Icon sdsIcon="xMark" sdsSize="l" sdsType="iconButton" />,
-  sdsSize: "medium",
-  sdsType: "tertiary",
+  sdsType: "secondary",
 };
 
-export const SmallPrimaryIconButton = Template.bind({});
+const LivePreviewDemo = (props: Args): JSX.Element => {
+  const spacings = getSpaces(props);
```
masoudmanson, line 59: Since we have the `defaultTheme` object, we can retrieve `spaces` from it directly.

```
const spacings = defaultAppTheme?.spacing;
```

src/core/IconButton/index.stories.tsx
```diff
@@ -28,72 +29,158 @@ export default {
...
-};
-
-export const SmallSecondaryIconButton = Template.bind({});
+  const livePreviewStyles = {
+    alignItems: "center",
+    display: "grid",
+    gridColumnGap: "0px",
+    gridRowGap: "50px",
+    gridTemplateColumns: "repeat(11, min-content)",
+  };
```
masoudmanson, line 67: I think we can change the grid styling, and put every couple of icons inside one separate div and give them marginRight with inline styles.

```
const livePreviewStyles = {
    alignItems: "center",
    display: "grid",
    gridColumnGap: 24,
    gridTemplateColumns: "repeat(11, min-content)",
};
```

src/core/IconButton/index.stories.tsx
```diff
@@ -28,72 +29,158 @@ export default {
...
 
-SmallSecondaryIconButton.args = {
-  "aria-label": "checked",
-  disabled: false,
-  icon: <Icon sdsIcon="checkCircle" sdsSize="s" sdsType="iconButton" />,
-  sdsSize: "small",
-  sdsType: "secondary",
+  return (
+    <div style={livePreviewStyles as React.CSSProperties}>
+      <div>
```
masoudmanson, line 71: Merge these divs so that the 24px gap between the columns would be ok.

```
<div style={{ display: "flex" }}>
  <Demo
    style={{ marginRight: spacings?.xxs }}
    icon={<Icon sdsIcon="grid" sdsSize="l" sdsType="iconButton" />}
    sdsSize="large"
    sdsType="primary"
  />

  <Demo
    style={{ marginRight: spacings?.xxs }}
    icon={<Icon sdsIcon="grid" sdsSize="l" sdsType="iconButton" />}
    sdsSize="large"
    sdsType="primary"
  />
</div>
```

src/core/IconButton/index.stories.tsx
```diff
@@ -28,72 +29,158 @@ export default {
...
+      </div>
+      <div>
+        <Demo
+          style={{ marginRight: spacings?.xxs }}
+          icon={<Icon sdsIcon="grid" sdsSize="l" sdsType="iconButton" />}
+          sdsSize="large"
+          sdsType="primary"
+        />
+      </div>
+      <div>
```
masoudmanson, line 87: We can merge these two divs as well.

src/core/IconButton/index.stories.tsx
```diff
@@ -28,72 +29,158 @@ export default {
...
+        <Demo
+          style={{ marginLeft: "24px", marginRight: "10px" }}
```
masoudmanson, line 89: Since we have the theme spacings, we can use it here. And since we have gap between columns, there is no need for marginLeft.

`style={{ marginRight: spacings?.m }}`

You can apply this change to the following components too.

masoudmanson (approved):
Thanks @bethbertozzi 
It looks good to me 💯

src/core/IconButton/index.stories.tsx
bethbertozzi: @masoudmanson I used spacings originally but it was not working for some reason. I will try it again

PR #217, opened 2022-07-12
refactor(radio button): change stories to csf format

bethbertozzi:
## Summary

Shortcut ticket: [sh-XXXX](https://app.shortcut.com/sci-design-system/story/167634/update-stories-for-inputradio-to-csf-format)

The caption functionality is not yet implemented, I have made a ticket to implement it here: https://app.shortcut.com/sci-design-system/story/205655/add-caption-to-radio-button-component
As well, the label is not currently part of the library component, I have created a ticket to add the label to the component (consistent with checkbox) here: https://app.shortcut.com/sci-design-system/story/205654/add-optional-label-to-radio-button-component

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/src/core/Radio/index.stories.tsx
+++ b/src/core/Radio/index.stories.tsx
@@ -13,7 +13,7 @@ const DefaultDemo = (props: Args): JSX.Element => {
       name="radio-buttons-group"
     >
       <FormControlLabel
-        control={<RadioButton stage="checked" />}
+        control={<RadioButton stage="checked" {...props} />}
         label={text}
         value="demo"
       />
@@ -43,7 +43,7 @@ Default.args = {
 const LivePreviewDemo = (props: Args): JSX.Element => {
   const { text } = props;
 
-  const [checked, setChecked] = React.useState([true, false]);
+  const [checked, setChecked] = React.useState([true, false, false]);
 
   const handleCheck1 = () => {
     setChecked([true, false, false]);
@@ -63,6 +63,7 @@ const LivePreviewDemo = (props: Args): JSX.Element => {
             <RadioButton
               onChange={handleCheck1}
               stage={checked[0] ? "checked" : "unchecked"}
+              data-testid="radioButton"
             />
           }
           label={text}
@@ -75,6 +76,7 @@ const LivePreviewDemo = (props: Args): JSX.Element => {
             <RadioButton
               stage={checked[1] ? "checked" : "unchecked"}
               onChange={handleCheck2}
+              data-testid="radioButton"
             />
           }
           label={text}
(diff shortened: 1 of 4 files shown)


src/core/Radio/index.tsx
```diff
@@ -37,6 +37,7 @@ const RadioButton = (props: RadioProps): JSX.Element => {
 
   return (
     <StyledRadioButton
+      data-testid="radioButton"
```
masoudmanson, line 40: I think you should not add a `data-testid` here, in the main component code! 
You can add it to the story function and use it! 
You can take a look at the `InputSlider` component for an example. In the `index.stories.tsx` you'll find a `TestDemo` function which returns a JSX element with a `data-testid`.

src/core/Radio/index.stories.tsx
```diff
@@ -1,77 +1,117 @@
...
+const DefaultDemo = (props: Args): JSX.Element => {
+  const { text } = props;
+
+  return (
+    <RadioGroup
+      aria-labelledby="demo-radio-buttons-group-label"
+      defaultValue="demo"
+      name="radio-buttons-group"
+    >
+      <FormControlLabel
```
masoudmanson, line 15: It is a good practice to pass the `{...props}` to the component so that the storybook controls could change the component! 
Currently, one can't disable the radio input for example!

src/core/Radio/index.stories.tsx
```diff
@@ -1,77 +1,117 @@
...
-      inputProps={{ "aria-label": "checked and disabled" }}
-      onChange={actions.onChange}
-      stage="checked"
-    />
-  </div>
-));
+const LivePreviewDemo = (props: Args): JSX.Element => {
+  const { text } = props;
+
+  const [checked, setChecked] = React.useState([true, false]);
```
masoudmanson, line 46: You have initialized the state with a tuple of `[boolean, boolean]`, but down there, in the `handleCheck` functions, you are setting an array of three booleans to the state!

masoudmanson (commented):
Thanks for another well written PR @bethbertozzi

clarsen-czi (changes requested):
Looking good, @bethbertozzi ! Just a few small tweaks to spacing and we should be golden.

masoudmanson (approved):
LGTM 💯  Thank you @bethbertozzi  🤩

bethbertozzi:
The space between the label and the button will be adjusted in [this ticket](https://app.shortcut.com/sci-design-system/story/205654/add-optional-label-to-radio-button-component), when the label is added to the component

clarsen-czi (approved):
LGTM! Thanks @bethbertozzi !!

PR #218, opened 2022-07-18
refactor(loading indicator): refactor stories to CSF

bethbertozzi:
## Summary

**Loading Indicator**
Shortcut ticket: [sh-167629](https://app.shortcut.com/sci-design-system/story/167629/update-stories-for-loadingindicator-to-csf-format)

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


clarsen-czi (approved):
LGTM! Thanks @bethbertozzi !!

masoudmanson (approved):

PR #229, opened 2022-08-02
docs(readme): Update readme file

masoudmanson:
Update Readme file


jfoo1984 (approved):

README.md
```diff
@@ -212,8 +212,8 @@ extra theme variables based on the themeOptions provided, so if you override `de
 ```tsx
   import { ThemeProvider as EmotionThemeProvider } from "@emotion/react";
   import { defaultAppTheme, makeThemeOptions } from "czifui";
-  import { StylesProvider, ThemeProvider } from "@material-ui/core/styles";
-  import createTheme from "@material-ui/core/styles/createTheme";
+  import { StylesProvider, ThemeProvider } from "@mui/material/styles";
```
jfoo1984, line 215: For CZ ID, I actually had to drop `StylesProvider` to get this to work:
https://github.com/chanzuckerberg/czid-web-private/pull/2152/files#diff-65fab8f3bb9e258bee21db78d8bd1335a5a2d231a98c65f0667949512dfd7f94R72-R76

I don't know if that is specific to CZ ID, or something that we need to update here.  We weren't doing any styling using `StylesProvider` in the CZ ID app though, so it was easy to drop.

tihuan: Thanks so much for raising this, @jfoo1984 🙏  I know we have a [Slack thread](https://czi-sci.slack.com/archives/C032S43KKFV/p1659143037297849) on this already, but for completeness I'll add a quick summary here:

1. We can use `StylesEngineProvider` over `StylesProvider` if needed
2. Originally `StylesProvider` is recommended, so we can add `injectFirst` prop that tells MUI to inject the default styles first, which allows project's custom CSS in JS styles to overwrite the default MUI style. We could make adding `<StylesEngineProvider injectFirst />` optional for projects in SDS README, the only concern I can think of is that if the project goes with CSS in JS solution down the road, they might not know/remember to add `<StylesEngineProvider injectFirst />`?

PR #234, opened 2022-08-11
feat(callout): add expandable variant

jfoo1984:
- Add logic to separate title from other child content to make a callout expandable
- Add story for new variant

## Summary

Shortcut ticket: [sh-155242](https://app.shortcut.com/sci-design-system/story/155242/add-expandable-variant-to-callout-component)

Implement the Expandable Callout (a type of callout) component.

## Checklist

- [x] Test Story in Storybook
- [ ] Tests written


src/core/Callout/index.tsx
```diff
@@ -59,29 +66,66 @@ const Callout = ({
...
+        sdsType="tertiary"
+        size="large"
+      >
+        {" "}
+        <Icon sdsIcon="xMark" sdsSize="s" sdsType="iconButton" />{" "}
+      </ButtonIcon>
+    ) : null;
+  };
+
+  const expanded = !expandable || stage === "open";
```
mmmmmmaya, line 100: why not use the constant here?

src/core/Callout/index.tsx
```diff
@@ -59,29 +66,66 @@ const Callout = ({
...
+  let calloutContent;
+
+  if (Array.isArray(children)) {
+    if (children[0].type.name === "CalloutTitle") {
+      [calloutTitle, ...calloutContent] = children;
+    } else {
+      // no CalloutTitle component, so use all child nodes as title
+      calloutTitle = children;
+    }
+  } else {
```
mmmmmmaya, line 112: both else statements results in the same action. wondering if we can combine the conditionals into a single if-statement (could stored each conditional in a named var first to make it readable if you want)?

src/core/Callout/index.tsx
```diff
@@ -59,29 +66,66 @@ const Callout = ({
...
+    ) : null;
+  };
+
+  const expanded = !expandable || stage === "open";
+
+  let calloutTitle = children;
+  let calloutContent;
+
+  if (Array.isArray(children)) {
+    if (children[0].type.name === "CalloutTitle") {
```
mmmmmmaya, line 106: would you need fallbacks here in case children is empty? for example `children[0]?.type.name`. not sure if that's a state you can get in.

clarsen-czi (changes requested):
See comment in Chromatic re: bottom margin on the collapsed view of collapsible variant.

src/core/Callout/index.tsx
jfoo1984: yea I think that makes sense.  I'm not sure how likely that state is either, but it seems like it doesn't hurt.  I'll also add some notes to the PR description on how these conditions work in terms  of what is shown based on what the user provides.

jfoo1984: yea let me fix this up a bit.  I think I actually don't need either of the `calloutTitle = children`, because that's already assigned above.  I can just move the explanatory comments up to that variable initialization.

src/core/Callout/index.tsx
jfoo1984: yea let me fix this up a bit.  I think I actually don't need either of the `calloutTitle = children`, because that's already assigned above.  I can just move the explanatory comments up to that variable initialization.

jfoo1984:
@clarsen-czi and @mmmmmmaya, sorry for the delay, but I think I've address both of your comments on this PR!  Let me know how it looks

jfoo1984:
Actually @clarsen-czi, I noticed one other quirk.  mui also includes top margin for the title, which gets added when I expand the callout.  so I think I need to remove that top margin as well.

clarsen-czi (changes requested):
Everything is looking great! Just one small change to the widths of the instances in the live preview view and we should be good. Thanks @jfoo1984 !!

masoudmanson (approved):
Thank you @jfoo1984, this looks amazing :D
LGTM 💯

jfoo1984:
@clarsen-czi I've updated the live preview based on your feedback!  would you be able to take a look again when you have a ahcne?

clarsen-czi (approved):
Looks great @jfoo1984 ! Thanks for the quick fix.

PR #236, opened 2022-08-12
fix(link): type

tihuan:
I removed manual casting SDS Link as `RawLink`, since that's the root cause of the bug @ehoops-cz 's seeing in [slack](https://czi-sci.slack.com/archives/C032S43KKFV/p1660325285598779)

The intention was to fix a Link linter error in [this commit](https://github.com/chanzuckerberg/sci-components/commit/bfbb2f23cfc22b3b72e047df528cbe112848bdce), which I think casting the `styled(Link) as typeof Link` should have been enough? @masoudmanson please help confirm on Monday 🙏 Thanks so much!

Screen Shot 2022-08-12 at 2 14 27 PM

### DIFF
--- a/src/core/Link/index.tsx
+++ b/src/core/Link/index.tsx
@@ -1,4 +1,3 @@
-import { Link as RawLink } from "@mui/material";
 import React, { ForwardedRef, forwardRef } from "react";
 import { LinkProps, StyledLink } from "./style";
 
@@ -6,7 +5,10 @@ import { LinkProps, StyledLink } from "./style";
  * @see https://v4.mui.com/components/links/
  */
 const Link = forwardRef(
-  (props: LinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
+  <C extends React.ElementType>(
+    props: LinkProps<C>,
+    ref: ForwardedRef<HTMLAnchorElement>
+  ) => {
     const { sdsStyle } = props;
     let underline: LinkProps["underline"];
 
@@ -16,7 +18,7 @@ const Link = forwardRef(
 
     return <StyledLink {...props} underline={underline} ref={ref} />;
   }
-) as typeof RawLink;
+);
 
 export type { LinkProps };
 
--- a/src/common/styles-dictionary/css/variables.css
+++ b/src/common/styles-dictionary/css/variables.css
@@ -1,6 +1,6 @@
 /**
  * Do not edit directly
- * Generated on Fri, 12 Aug 2022 18:51:53 GMT
+ * Generated on Fri, 12 Aug 2022 21:49:52 GMT
  */
 
 :root {
--- a/src/common/styles-dictionary/scss/_variables.scss
+++ b/src/common/styles-dictionary/scss/_variables.scss
@@ -1,6 +1,6 @@
 
 // Do not edit directly
-// Generated on Fri, 12 Aug 2022 18:51:53 GMT
+// Generated on Fri, 12 Aug 2022 21:49:52 GMT
 
 $sds-font-letter-spacing-default: 0.3px;
 $sds-font-letter-spacing-caps: 1px;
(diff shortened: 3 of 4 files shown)


src/core/Link/style.ts
```diff
@@ -3,12 +3,7 @@ import { Link, LinkProps as RawLinkProps } from "@mui/material";
 import { styled } from "@mui/material/styles";
 import { CommonThemeProps as StyleProps, getBorders } from "../styles";
 
-// (thuang): Support `component` prop
-// https://stackoverflow.com/a/66123108
-export type LinkProps<C extends React.ElementType = "a"> = RawLinkProps<
-  C,
-  { component?: C }
```
ehoops-cz, line 10: do we want to remove this?  I'm curious if this would let us use a `button` instead when appropriate - the one I have has an onClick attached, so should probably be a `button` html element.

tihuan: Ooh you're right 😆 Just tested it locally and also needed to add a bit more type in Link component to make `component` prop work properly 👌 

And you're right that since there's onClick and it doesn't cause a page navigation, it should be a button element!

PTAL again thank you!

src/core/Link/index.tsx
```diff
@@ -1,12 +1,14 @@
-import { Link as RawLink } from "@mui/material";
 import React, { ForwardedRef, forwardRef } from "react";
 import { LinkProps, StyledLink } from "./style";
 
 /**
  * @see https://v4.mui.com/components/links/
  */
 const Link = forwardRef(
-  (props: LinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
+  <C extends React.ElementType>(
```
tihuan, line 8: this is also needed for supporting `component` prop in addition to:

```ts
// (thuang): Support `component` prop
// https://stackoverflow.com/a/66123108
export type LinkProps<C extends React.ElementType = "a"> = RawLinkProps<
  C,
  { component?: C }
> &
  StyleProps & {
    sdsStyle?: "default" | "dashed";
  };
```

tihuan: Maybe @masoudmanson was trying to solve the `component` prop issue with casting Link as `RawLink`?

ehoops-cz (approved):
Looks good to me!  Thanks!

tihuan:
@masoudmanson PTAL when you get a chance 🙏 Thanks so much!

tihuan:
looks like Masoud is out today, hopefully it's not blocking you today, @ehoops-cz ? Thanks!

src/core/Link/index.tsx
masoudmanson: Thank you @tihuan for fixing this one, There was a linter error on the link component and I had to export it as the type of Link back then. Didn't know it would cause problems later. 💯

masoudmanson (approved):
Awesome, Thank you @tihuan 🥇

masoudmanson:
> @masoudmanson PTAL when you get a chance 🙏 Thanks so much!

The code is completely OK 🤩 , Thanks again 🙏🏻
I think we also need @clarsen-czi 's approval for the live preview section.

tihuan:
Thanks so much, @masoudmanson !

@clarsen-czi PTAL the following storybook change 🙏 I added a new section to test that the Link component can be rendered as a button element without any type errors too!

https://www.chromatic.com/test?appId=616981d9028812003ade73f3&id=62f6cd23fc5838125eb2507f

Screen Shot 2022-08-15 at 2 15 12 PM

tihuan:
Added @chanzuckerberg/sds-design for storybook addition!

clarsen-czi (approved):
LGTM! Thanks all!!

tihuan:
Awesomeee thanks so much, @clarsen-czi 🎉 !

PR #249, opened 2022-08-31
refactor(cellheader): masoudamjadi/update textPosition to horizontalAlign

masoudmanson:
## Summary

**CellHeader**
Change `textPosition` prop name to `horizontalAlign`.


clarsen-czi (dismissed):
Double check that the direction prop in the test view is working as intended; otherwise looks great @masoudmanson !

masoudmanson:
> Double check that the direction prop in the test view is working as intended; otherwise looks great @masoudmanson !

The test view won't take dynamic props and its props are hard coded. I will remove the storybook controls for the test view as well.

clarsen-czi (approved):
LGTM!

PR #261, opened 2022-09-14
feat(table): table horizontal scrolling + Tag style fix

masoudmanson:
Horizontal Scrolling for the Table Component + Some style fixes


src/core/CellComponent/index.tsx
```diff
@@ -14,7 +14,9 @@ const CellComponent = forwardRef(
     const { children } = props;
 
     return (
-      <StyledCellComponentData {...props}>{children}</StyledCellComponentData>
+      <StyledCellComponentData className="cell-component" {...props}>
```
bethbertozzi, line 17: thanks for catching this

masoudmanson: When the TableRow is disabled, I needed this className to properly style the cell components!

clarsen-czi (changes requested):
Looking good! Just a small change to make on the margin on the TagFilter component and then we should be good.

clarsen-czi (approved):
Looks good!

PR #262, opened 2022-09-15
feat(cellbasic): add icon support to CellBasic

masoudmanson:


src/core/CellBasic/index.stories.tsx
```diff
@@ -25,12 +26,34 @@ const CellBasic = (props: Args): JSX.Element => {
...
+          "XS Download",
+          "S Light bulb",
+          "L Bacteria",
+          "XL Flask",
+        ],
+        type: "select",
+      },
+      mapping: availableIconOptions,
+      options: Object.keys(availableIconOptions),
+    },
```
bethbertozzi, line 56: add icon as control to exclude in test

bethbertozzi (changes requested):
I think it would be good to have a test with the icon, but other than that and a small cleanup it looks really great 😃 thanks Masoud!

src/core/CellBasic/index.test.tsx
```diff
@@ -43,4 +43,11 @@ describe("<CellBasic />", () => {
     const style = window.getComputedStyle(renderedElement);
     expect(style.verticalAlign).toBe("bottom");
   });
+
```
bethbertozzi, line 46: 🙌

bethbertozzi (approved):
looks great!

clarsen-czi (approved):
LGTM! Thanks @masoudmanson !!

PR #270, opened 2022-10-06
[DO NOT MERGE] Example of dropdown onchange bug

jfoo1984:
## Summary
Slack context:
https://czi-sci.slack.com/archives/C032S43KKFV/p1665076950918139

> Issue Description: I'm not sure if this is a bug or intended behavior, but I noticed that if you have a parent component that included a `Dropdown`, and one of the props on the parent changes (even a prop that is not passed into the `Dropdown` at all).  It causes the dropdown to fire its `onChange` callback.  So basically that callback is being called even when the value selected is not changing.
I put together branch / PR demoing this (also included a screen recording in the description: https://github.com/chanzuckerberg/sci-components/pull/270 

Reproduce potential dropdown bug in storybook

### Video demo
https://user-images.githubusercontent.com/2072627/194223813-4eea18b7-2f71-41f5-baed-b9c372294a87.mov

## Checklist
- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/src/core/Dropdown/index.stories.tsx
+++ b/src/core/Dropdown/index.stories.tsx
@@ -21,7 +21,6 @@ const BugWrapper = (): JSX.Element => {
   const [buttonProp, setButtonProp] = useState("a");
   const clickHandler = () => {
     setButtonProp(`${buttonProp}a`);
-    console.log(buttonProp);
   };
   return (
     <>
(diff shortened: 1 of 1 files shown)


tihuan:
Thanks so much for the code example for reproducing the issue, Jerry! 

The bug has been fixed here 🙏 

https://github.com/chanzuckerberg/sci-components/pull/304

PTAL thank you!

PR #275, opened 2022-10-17
refactor(inputs): moved input stories into inputs folder in storybook

bethbertozzi:
## Summary

**InputCheckbox**
Shortcut ticket: [sh-XXXX](https://app.shortcut.com/sci-design-system/story/210952/adjust-name-of-checkbox-to-inputcheckbox-in-storybook)
Currently the InputCheckbox component is named Checkbox in Storybook, but should be named InputCheckbox

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


clarsen-czi (changes requested):
Looking good! I really like the idea of consolidating all of the inputs into their own group in Storybook, though make sure the prefix "Input-" is still in the names of the components since that's how they're named in our documentation and in the Figma libraries. One additional comment in there regarding the width of the InputDropdown when its being shown as part of a Dialog

bethbertozzi:
@clarsen-czi I forgot I can't see your comments after I add a commit (I think?), do you remember what the 'width of the InputDropdown when it's being shown as part of a Dialog' problem is? otherwise fixed those names!

bethbertozzi:
@clarsen-czi thanks! It looks like a quick fix I can work on next week. For now I added this ticket in fall cleaning for it: https://app.shortcut.com/sci-design-system/story/225821/fix-dropdown-in-dialog-width

clarsen-czi (approved):
Ok sounds good! Just make sure that the fix is just in that specific case and isn't necessarily a rule for all Dialogs (since some of them can get super wide; ultimately we want the design to dictate the width of the Input)

masoudmanson (approved):
Since you have updated the name of the `Checkbox` component to `InputCheckbox`, we should announce a breaking change in the channel. Thanks for the amazing PR @bethbertozzi 👍🏻

PR #277, opened 2022-11-01
feat(banner): remove text prop and add children

bethbertozzi:
BREAKING CHANGE: the banner component no longer accepts a text prop. This can be fixed by changing the text to be a child of the banner component

## Summary

**Banner**
Shortcut ticket: [sh-226199](https://app.shortcut.com/sci-design-system/story/226199/add-child-to-banner)
CZ Gen Epi wants to add a banner that includes a link. Currently the banner requires a text prop and does not accept children. As a result it isn't possible to include a link. Would SDS be open to a PR that would make it possible to pass a child component instead of text?

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/src/core/Banner/index.tsx
+++ b/src/core/Banner/index.tsx
@@ -6,28 +6,20 @@ import {
   IconWrapper,
   StyledBanner,
   StyledButtonIcon,
-  Text,
 } from "./style";
 
 export interface BannerProps extends BannerExtraProps {
+  children: React.ReactNode;
   dismissed?: boolean;
   dismissible?: boolean;
   onClose?: (e: React.MouseEvent) => void;
-  text: string;
 }
 
 const Banner = forwardRef<HTMLDivElement, BannerProps>(function Banner(
   props,
   ref
 ): JSX.Element | null {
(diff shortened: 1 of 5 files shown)


src/core/Banner/index.tsx
```diff
@@ -48,6 +49,7 @@ const Banner = forwardRef<HTMLDivElement, BannerProps>(function Banner(
           <Icon sdsIcon="infoCircle" sdsSize="l" sdsType="static" />
         </IconWrapper>
         <Text>{text}</Text>
+        {props.children}
```
ehoops-cz, line 52: This leaves some confusion around what will get rendered when text and children are passed in. It would be more clear to throw an error if the user passes both `text` and `children` props.

ehoops-cz (approved):
This works for us.  One comment re: making the behavior a bit more clear.

masoudmanson (approved):
LGTM, Thanks Beth

clarsen-czi (approved):
Looks great! Thanks @bethbertozzi !

PR #291, opened 2022-11-18
feat(cellheader): allow more than strings as CellHeader children

mmmmmmaya:
Allow components as children of CellHeader (necessary for a checkbox in header row, for example)


masoudmanson (approved):
LGTM, Thanks @mmmmmmaya

PR #294, opened 2022-11-27
fix(dropdown): change the width of the dropdown inside a modal to 100%

masoudmanson:
## Summary

**Dropdown**
Shortcut ticket: [sh-225821](https://app.shortcut.com/sci-design-system/story/225821/fix-dropdown-in-dialog-width)
The issue with the dropdown input when it's in a dialog is that it should span the width of the dialog and its width should be determined by the padding on the dialog (at least in this specific case; it just visually looks strange being shorter like it is)

Note: This is specific to the InputDropdown example in Storybook where the input is placed inside a Dialog; NOT a rule for all dialogs

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


clarsen-czi (approved):
LGTM! Thanks @masoudmanson !!

PR #295, opened 2022-11-27
refactor(accordion): add forwardRef to Accordion + Fix data-testId

masoudmanson:
## Summary

Shortcut ticket: [sh-224676](https://app.shortcut.com/sci-design-system/story/224676/remove-data-testid-from-index-files)

Right now, data-testid is sometimes included in the index file of components and is then always present whenever that component is used. It should only be used in the test file. We need to remove it from index files and add it to test files where needed.

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/src/core/Accordion/index.stories.tsx
+++ b/src/core/Accordion/index.stories.tsx
@@ -2,12 +2,17 @@ import { Args, Story } from "@storybook/react";
 import React from "react";
 import AccordionDetails from "./components/AccordionDetails";
 import AccordionHeader from "./components/AccordionHeader";
-import Accordion from "./index";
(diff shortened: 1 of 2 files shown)


bethbertozzi (approved):
looks great! thank you masoud. I wonder if we could add this as a lint rule, like data-testid is only allowed in certain file types

masoudmanson:
> looks great! thank you masoud. I wonder if we could add this as a lint rule, like data-testid is only allowed in certain file types

Hmm, that sounds like a good idea. I'll look into it :D
Thank you @bethbertozzi 👍🏻

PR #308, opened 2023-01-03
fix(dropdownmenu): remove onClickAway from forwardedProps

masoudmanson:
## 2023's first commit   🎉 ✨ 🥳

**DropdownMenu**

Remove `onClickAway` from forwardedProps

### DIFF
--- a/src/core/ComplexFilter/index.tsx
+++ b/src/core/ComplexFilter/index.tsx
@@ -33,24 +33,20 @@ export interface ComplexFilterProps<Multiple> {
 }
 
 // eslint-disable-next-line sonarjs/cognitive-complexity
-const ComplexFilter = <Multiple extends boolean | undefined = false>(
-  props: ComplexFilterProps<Multiple>
-): JSX.Element => {
-  const {
-    options,
-    label = "",
-    multiple = false,
-    search = false,
-    onChange,
-    DropdownMenuProps = {},
-    InputDropdownProps = { sdsStyle: "minimal" },
-    value: propValue,
-    PopperComponent,
-    InputDropdownComponent = InputDropdown,
-    isTriggerChangeOnOptionClick = false,
-    ...rest
-  } = props;
-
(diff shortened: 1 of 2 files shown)


src/core/ComplexFilter/index.tsx
```diff
@@ -71,50 +75,17 @@ const ComplexFilter = <Multiple extends boolean | undefined = false>({
...
-          onClick={handleClick}
-          sdsStage={open ? "userInput" : "default"}
-          {...InputDropdownProps}
-        />
-
-        <StyledChipsWrapper>
-          <Chips value={value} multiple={multiple} onDelete={handleDelete} />
-        </StyledChipsWrapper>
-      </Wrapper>
+  const handleClose = () => {
```
tihuan, line 78: I see that the handler functions are changed from function declaration to anonymous function expression, and moved above the return statement

Just wanted to provide some context for the code structure there 😄 --

Traditionally people prefer to declare functions before using them, but nowadays with IDEs being able to jump to function definitions easily, I've leaned towards a newer practice of putting what the component returns above the fold (what people see in IDE without scrolling down), since that's most readers care about the most, and hide the implementation details below the fold, so readers only need to scroll to find the implementation if needed

Also function declaration helps with error stack trace, since the functions aren't anonymous!

Please let me know what you think of this rationale 🙏 Thank you!

tihuan (approved):
Woohoo! Thanks so much for the first PR of the year, @masoudmanson 🤩 !

Just a quick comment on the code structure refactoring! Thanks again!

src/core/ComplexFilter/index.tsx
masoudmanson: Oh, that sounds just right. I totally agree with that. I'll revert the unnecessary changes. Thank you 🫠

tihuan: Awesome! So glad you like the rationale too 🤩 👏 I still appreciate the effort you took to refactor the code even though we're not using it this time 🙏

PR #309, opened 2023-01-09
feat(demos): testing the migration of CodeSandbox demos to SDS git repo

masoudmanson:
👀 Adding one Table demo to the SDS git repo to test its usage on the CodeSandbox.


tihuan (approved):
Awesomee thanks so much for testing this out, Masoud! Hope it works 🤞

PR #310, opened 2023-01-10
fix(table demos): Masoudamjadi/codesandbox demos

masoudmanson:
Fix an infinite loop bug in `makeData.ts` file 😱


demos/TableWithFrontEndPagination/src/makeData.ts
```diff
@@ -12,7 +12,8 @@ export type Person = {
 
 const range = (len: number) => {
   const arr: number[] = [];
-  for (let i = 0; i < len; i + 1) {
+  // eslint-disable-next-line no-plusplus
+  for (let i = 0; i < len; i++) {
```
tihuan, line 16: For range, you could do Array.from(Array(len))!

PR #312, opened 2023-01-17
style(buttonicon): fix primary buttonIcon size

masoudmanson:
A 'box-sizing' of 'content-box' resolved the primary ButtonIcon's size issue.

### DIFF
--- a/src/core/ButtonIcon/style.ts
+++ b/src/core/ButtonIcon/style.ts
@@ -152,6 +152,7 @@ const large = <ButtonIconSize extends keyof ButtonIconSizeToTypes>(
 
   return css`
     .MuiSvgIcon-root {
+      box-sizing: content-box;
       height: ${iconSizes?.xl.height}px;
       width: ${iconSizes?.xl.height}px;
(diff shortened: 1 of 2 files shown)


src/core/ButtonIcon/style.ts
```diff
@@ -152,6 +152,7 @@ const large = <ButtonIconSize extends keyof ButtonIconSizeToTypes>(
 
   return css`
     .MuiSvgIcon-root {
+      box-sizing: content-box;
```
tihuan, line 155: can we add a comment above to explain why we use content-box here instead of the default border-box please 🙏 Thanks so much!

tihuan (approved):
Awesomee v elegant solution! Thanks so much, Masoud ✨ Just one quick comment!

src/core/ButtonIcon/style.ts
masoudmanson: Sure, I'll add some 🤓

masoudmanson: Eew, we don't even need to adjust the box size if we simply replaced padding with margin. 🫠 My bad 😵‍💫

PR #313, opened 2023-01-17
fix(buttonicon): remove sdsIconProps from forwardedProps

masoudmanson:
Remove `SdsIconProps` from forwardedProps in ButtonIcon component.


tihuan (approved):
Great catch thank you!

Issue #41, opened 2021-08-03
ButtonIcon onFocus behavior


ehoops-zz:
The current buttons do not show the default focus border.  This might be an MUI issue, but in any case, we should add it back to make accessibility the default.

I used this to add it back in the Rare As One code:
```
export const StyledButton = styled(Button)`
  :focus {
    outline: 5px auto Highlight;
    outline: 5px auto -webkit-focus-ring-color;
  }
```

PR #319, opened 2023-01-19
style(buttonicon): add onFocus highlight border

masoudmanson:
#41


src/core/ButtonIcon/style.ts
```diff
@@ -172,6 +172,11 @@ export const StyledButtonIcon = styled(IconButton, {
 })`
   padding: 0;
 
+  :focus {
+    outline: 5px auto Highlight;
```
tihuan, line 176: ooh `highlight` is convenient 💡 How is the color value calculated? Is it based on whatever the browser defines?

tihuan (approved):
Nice solution! Thanks so much, Masoud 🙏 ✨

src/core/ButtonIcon/style.ts
masoudmanson: Yes, exactly. `Highlight` is the default value for focus colors in browsers like Firefox and Safari. Chrome, on the other hand, has specified its own value as `-webkit-focus-ring-color`, which is different in color but achieves the same thing.

Issue #337, opened 2023-01-30
New single-select InputDropdown variant


cathystoli:
**Component Name:** InputDropdown
**Component Requirements:** - New component variation, specific parameters as below
- sdsType: single-select
- sdsStyle: minimal
- Details: on

**Design specs:** https://www.figma.com/file/RAIg2Gzsw2QylzY28LYsl5/InputDropdown-Exploration?node-id=55%3A539&t=ZFHEOb3IWQMPkJqf-1
**Team(s) Requesting:** CZID
**Component Needed By:** This is part of the Heatmap - Move Filters to the left project. I will share the eng point of contact after handoff

**Plan Details:**
- There is a version of this that exists but does not match the full requirements, we will take what exists in the code and updating it 

- [x] #343 blocker for this ticket 
- [ ] Eng implementation - @liaprins this ticket

cathystoli:
@clarsen-czi can you review the request above? It would be good to get your thoughts on the broader use for this change

clarsen-czi:
@cathystoli Kev and I chatted about this one before he posted to the SDS request channel and I let him know I thought it would make a good addition to SDS.

For context, it would be an additional variant of the minimal dropdown input allowing for the ability to include text with additional details as part of the input (such as what or how many options have been selected in the accompanying dropdown menu). Currently the rounded and square dropdown inputs have this ability, but the minimal version does not, so this is just expanding that ability to the minimal version.

LMK if you're looking for any more details about broader implications!

liaprins:
> ability to include text with additional details as part of the input... (such as 
> 1. what or 
> 2. how many options have been selected in the accompanying dropdown menu

@clarsen-czi Do the different examples you listed above (I split into numbered list of "1" and "2" above) need to each be a distinct variant? Or there just needs to be one variant that can accommodate text in general? If the former, are there any other usecases that would warrant their own variant for this component as well (in addition to 1 and 2 above)? Thanks!

tihuan:
@clarsen-czi 

Some quick questions 😄 🙏 

1. In the [design](https://www.figma.com/file/v8T3vabzkmbvH71sQ1EXzh/Kevin---Heatmap-Filters-Redesign?node-id=1047%3A35901&t=6w4yzPlbDqUARD6f-1), the minimal InputDropdown has the detail/count in a new line, is that always the case or only when they can't fit on one line?

[...]

PR #354, opened 2023-02-10
feat(minimal inputdropdown): update Minimal Dropdown Style

tihuan:
## Summary

Ticket: #337 

Component Name: InputDropdown
Component Requirements: - New component variation, specific parameters as below

sdsType: single-select
sdsStyle: minimal
Details: on
Design link: https://www.figma.com/file/v8T3vabzkmbvH71sQ1EXzh/Kevin---Heatmap-Filters-Redesign?node-id=1047%3A35901&t=8JY2ukyrEk5NnJ6M-1
Team(s) Requesting: CZID
Component Needed By: This is part of the Heatmap - Move Filters to the left project. I will share the eng point of contact after handoff

Plan Details:

There is a version of this that exists but does not match the full requirements, we will take what exists in the code and updating it

## Checklist

- [x] Test Story in Storybook
- [x] Tests written

### DIFF
--- a/src/core/Accordion/__snapshots__/index.test.tsx.snap
+++ b/src/core/Accordion/__snapshots__/index.test.tsx.snap
@@ -23,7 +23,7 @@ exports[`<Accordion /> Test story renders snapshot 1`] = `
       class="MuiAccordionSummary-expandIconWrapper css-yw020d-MuiAccordionSummary-expandIconWrapper"
     >
       <div
-        class="css-jbvua0"
+        class="css-1ft6wgl"
       >
         <svg
           aria-hidden="true"
--- a/src/core/Icon/style.ts
+++ b/src/core/Icon/style.ts
@@ -91,5 +91,6 @@ export const StyledSvgIcon = styled(SvgIcon, {
 `;
 
 export const StyledIcon = styled("div")`
+  /* display: contents causes an element's children to appear as if they were direct children of the element's parent, ignoring the element itself. This can be useful when a wrapper element should be ignored when using CSS grid or similar layout techniques. */
   display: contents;
 `;
--- a/src/core/InputDropdown/style.ts
+++ b/src/core/InputDropdown/style.ts
@@ -23,6 +23,7 @@ export interface InputDropdownProps extends CommonThemeProps {
   sdsType?: "singleSelect" | "multiSelect";
   details?: string;
   counter?: string;
+  shouldTruncateMinimalDetails?: boolean;
 }
 
 const labelFontBodyS = fontBody("s");
@@ -36,18 +37,17 @@ const inputDropdownStyles = (props: InputDropdownProps): SerializedStyles => {
 
   return css`
     border: ${borders?.gray[400]};
-    color: ${colors?.gray[500]};
+    color: black;
     cursor: pointer;
     padding: ${spacings?.xs}px;
+    /* minimal left right will be s px instead */
 
     &.MuiButton-text {
-      justify-content: flex-start;
-
       &:hover {
         color: #000;
       }
 
-      > span {
+      .styled-label {
         margin-right: ${spacings?.xs}px;
         margin-left: ${spacings?.xs}px;
         overflow: hidden;
@@ -74,7 +74,7 @@ const inputDropdownStyles = (props: InputDropdownProps): SerializedStyles => {
         fill: ${colors?.gray[600]};
       }
 
-      > span {
+      .styled-label {
         color: #000;
       }
     }
@@ -97,13 +97,22 @@ const inputDropdownStyles = (props: InputDropdownProps): SerializedStyles => {
 const minimal = (props: InputDropdownProps): SerializedStyles => {
   const colors = getColors(props);
   const palette = getPalette(props);
+  const spacings = getSpaces(props);
 
   return css`
-    border: none;
-    padding: 0;
+    ${labelStyle(props)}
 
-    & .MuiButton-label {
-      margin: 0;
+    align-items: flex-start;
+    border: none;
+    flex-direction: column;
+    padding: ${spacings?.xs}px ${spacings?.s}px;
+
+    /* Nesting to increase CSS specificity for style override */
+    &.MuiButton-text {
+      .styled-label {
+        margin: 0;
+        margin-right: ${spacings?.xs}px;
+      }
     }
 
     span {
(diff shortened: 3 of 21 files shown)


clarsen-czi (changes requested):
This looks great @liaprins !! Just a small tweak to the font style used in the minimal variant (it's a size larger than the rounded and square variants) and we should be good!

src/core/Icon/style.ts
```diff
@@ -91,5 +91,6 @@ export const StyledSvgIcon = styled(SvgIcon, {
 `;
 
 export const StyledIcon = styled("div")`
+  /* display: contents causes an element's children to appear as if they were direct children of the element's parent, ignoring the element itself. This can be useful when a wrapper element should be ignored when using CSS grid or similar layout techniques. */
```
masoudmanson, line 94: Thanks for the helpful comment 🙏🏻🤩

src/core/InputDropdown/style.ts
```diff
@@ -115,29 +123,13 @@ const minimal = (props: InputDropdownProps): SerializedStyles => {
     }
 
     &:hover {
-      color: ${colors?.gray[600]};
```
masoudmanson, line 118: Great job @liaprins, this looks much better now! 🥳

src/core/InputDropdown/style.ts
```diff
@@ -115,29 +123,13 @@ const minimal = (props: InputDropdownProps): SerializedStyles => {
...
-      span {
-        color: ${colors?.gray[600]};
-      }
-
-      path {
-        fill: ${colors?.gray[600]};
-      }
     }
 
     &:active {
```
masoudmanson, line 130: And here, thanks for clearing this up 😍

masoudmanson (approved):
It looks awesome 🤩 Thank you @liaprins and @tihuan 🥳😍

liaprins:
@clarsen-czi The label size should be updated now!

src/core/Accordion/__snapshots__/index.test.tsx.snap
```diff
@@ -23,7 +23,7 @@ exports[`<Accordion /> Test story renders snapshot 1`] = `
       class="MuiAccordionSummary-expandIconWrapper css-yw020d-MuiAccordionSummary-expandIconWrapper"
     >
       <div
-        class="css-jbvua0"
+        class="css-1ft6wgl"
```
liaprins, line 26: Expected that this CSS class name would change, because we updated its corresponding style component here and throughout this PR

src/core/InputDropdown/style.ts
```diff
@@ -115,37 +122,21 @@ const minimal = (props: InputDropdownProps): SerializedStyles => {
...
-      }
     }
 
     &:focus {
       outline: none;
     }
 
     &.MuiButton-root.MuiButton-text > span {
-      ${fontHeaderS(props)}
+      ${labelFontBodyS(props)}
```
liaprins, line 139: @masoudmanson We changed to `labelFontBodyS` because the Figma design file for the new minimal InputDropdown says to use fontBodyS

src/core/InputDropdown/style.ts
```diff
@@ -267,7 +249,13 @@ const doNotForwardProps = ["intent", "open", "sdsStage"];
 export const StyledInputDropdown = styled(Button, {
   shouldForwardProp: (prop: string) => !doNotForwardProps.includes(prop),
 })`
-  ${labelFontBodyS}
```
liaprins, line 270: We changed the default font body size style to be size xs (`labelFontBodyXs`) because all of the pieces of the inputDropdowns are size xs, _except_ for the minimal variant label and its details. So previously the label for all the InputDropdowns was being set to override this default (which had been `labelFontBodyS`) to be size xs. This in turn prevented the minimal variant's label from being able to override this.

So now the default is xs, and it doesn't have to be overridden by the label, details, and counter below. This also _does_ allow it to be overridden by just the minimal variant's label.

src/core/InputDropdown/style.ts
```diff
@@ -288,7 +276,6 @@ export const StyledInputDropdown = styled(Button, {
 export const StyledDetail = styled("span", {
   shouldForwardProp: (prop: string) => !doNotForwardProps.includes(prop),
 })`
-  ${labelFontBodyXs}
```
liaprins, line 291: Here's where we're removing the original override for all InputDropdowns to be size xs (now xs is the default which they inherit, so it doesn't need to be set here)

src/core/InputDropdown/style.ts
```diff
@@ -306,7 +293,6 @@ interface DetailsAndCounter extends CommonThemeProps {
 export const StyledLabel = styled("span", {
   shouldForwardProp: (prop: string) => !doNotForwardProps.includes(prop),
 })`
-  ${labelFontBodyXs}
```
liaprins, line 309: Here's where we're removing the original override for all InputDropdowns' counters to be size xs (now xs is the default which they inherit, so it doesn't need to be set here)

src/core/InputDropdown/style.ts
```diff
@@ -324,7 +310,6 @@ export const StyledLabel = styled("span", {
 export const StyledCounter = styled("span", {
   shouldForwardProp: (prop: string) => !doNotForwardProps.includes(prop),
 })`
-  ${labelFontBodyXs}
```
liaprins, line 327: Here's where we're removing the original override for all InputDropdowns' counters to be size xs (now xs is the default which they inherit, so it doesn't need to be set here)

src/core/InputDropdown/style.ts
```diff
@@ -338,3 +323,49 @@ export const StyledCounter = styled("span", {
   `;
   }}
 `;
+
+interface MinimalDetailsProps extends CommonThemeProps {
+  shouldTruncateMinimalDetails: InputDropdownProps["shouldTruncateMinimalDetails"];
+}
+
+export const MinimalDetails = styled("div")`
+  ${labelFontBodyS}
```
liaprins, line 332: This is now the only InputDropdown variant that needs to override the default xs font size to be size s instead (for the minimal variant's label).

The `interface` statement above lets the `labelFontBodyS` variable get the props it needs (@tihuan if I am correctly explaining?)

liaprins:
@masoudmanson We have made the update requested by Connor if you want to check! Thanks!

masoudmanson (approved):
Great work! I really appreciate the attention to detail in your code. @liaprins  🎉🤩

masoudmanson:
Before committing any code changes, it's important to run `yarn test -u` to update the previous test snapshots. The current error in GitHub checks (Push Tests failure) is likely occurring because the test snapshots are outdated. @liaprins

clarsen-czi (approved):
Looks great @liaprins! Thanks for the quick work 💪

Issue #336, opened 2023-01-26
Icon improvements - control over color


cathystoli:
**Component Name:** Icon

**Submitter**: ID, Jerry 

**Issue Description:** This is more of a suggestion or idea, but in CZ ID, we have one spot where style the flagCheck icon to be the  color green ( will attach a screenshot in thread).  I was wondering if it might make sense to provide more control over the color of the icon (rather than having to style them outside SDS). 
In this use case, the ability to specify the color of the icon as the "success, primary, or secondary" color would be sufficient, and those colors are already defined in the theme.  But a prop for color may be a possibility too, to allow other color options.

**How we will implement:** We will crate a prop for color so the teams are not limited, but they will be limited to the themes in each product library 

**Acceptance criteria:**

- [ ]

PR #358, opened 2023-02-15
feat(icon): add color prop to the icon component

masoudmanson:
## Summary

**Icon**
GitHub ticket: #336 

The  new color prop accepts one of the following values, which are defined in the defaultTheme.ts file:

- "beta"
- "gray"
- "primary"
- "secondary"
- "info"
- "success"
- "warning"
- "error"

However, these values can be modified through the use of custom themes created by products.

## Checklist

- [x] Test Story in Storybook
- [x] Tests written


src/core/Icon/index.stories.tsx
```diff
@@ -11,70 +11,71 @@ import {
...
   );
 };
 
 export default {
+  argTypes: {
+    color: {
+      control: {
+        type: "select",
+      },
+      options: [
```
tihuan, line 30: love the expanded options!

src/core/Icon/index.stories.tsx
```diff
@@ -92,10 +93,12 @@ const IconBankWrapper = styled("div")`
 `;
 
 const IconWrapper = styled("div")`
-  ${(props: CommonThemeProps) => {
+  ${(props: CommonThemeProps & { color?: SdsIconColorType }) => {
     const colors = getColors(props);
     const spacings = getSpaces(props);
 
+    const { color } = props;
```
tihuan, line 100: should we default this to `primary`?

E.g.,

```ts
const { color = "primary" } = props;

...

background-color: ${colors.[color][400]};
```

src/core/Icon/style.ts
```diff
@@ -35,19 +53,21 @@ function buttonStyle(): SerializedStyles {
 }
 
 function staticStyle<IconName extends keyof IconNameToSizes>(
-  props: IconExtraProps<IconName>
+  props: StyledSvgIconProps<IconName>
 ): SerializedStyles {
+  const { iconColor } = props;
```
tihuan, line 58: same here 😄 

```ts
const { iconColor = "primary" } = props;
```

src/core/Icon/style.ts
```diff
@@ -35,19 +53,21 @@ function buttonStyle(): SerializedStyles {
...
   const colors = getColors(props);
 
   return css`
-    color: ${colors?.primary[400]};
+    color: ${iconColor ? colors?.[iconColor][400] : colors?.primary[400]};
```
tihuan, line 62: `color: ${colors?.[iconColor][400]};`

tihuan (approved):
LGTM! Thanks so much, @masoudmanson 🙏 Just a tiny nit 😄 🥇

src/core/Icon/index.stories.tsx
masoudmanson: Aahh, that's much easier 😅

src/core/Icon/style.ts
masoudmanson: Thanks for the recommendation 🥳🤩

clarsen-czi (changes requested):
Conceptually the idea is there and works great! However I think we should provide more flexibility to folks so they can use any color from the library and aren't limited to the `-400` variants of them. See my comment in Chromatic for more details.

masoudmanson:
> Conceptually the idea is there and works great! However I think we should provide more flexibility to folks so they can use any color from the library and aren't limited to the `-400` variants of them. See my comment in Chromatic for more details.

A new prop named `shade` has been added to the `Icon` component, which can be used to set the color shade of the icon to 100, 200, 300, 400, 500 or 600.

tihuan (approved):
Great idea thanks both!

clarsen-czi (approved):
This looks great, thanks for adding the additional functionality, @masoudmanson !!

PR #362, opened 2023-02-21
fix(storybook): fix storybook UI styles

masoudmanson:
## Summary

**Storybook**
Shortcut ticket: #361


.storybook/manager-head.html
```diff
@@ -158,4 +172,58 @@
...
+      20
+    );
+    changeElementDimension(
+      document.querySelector('div[role="main"] > div > div ~ div'),
+      "height",
+      20
+    );
+  });
+
+  window.addEventListener(
```
tihuan, line 194: We can use ResizeObserver to offload some js work!

https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

tihuan (approved):
LGTM! Just a tiny perf improvement 🚀 Thanks so much!

.storybook/manager-head.html
masoudmanson: After implementing the feature using ResizeObservers, I came across this comment from the W3C technical architecture group about using Events and Observers appropriately, and I opted for a more traditional method. 🤔
What do you think? Should I change it to an observer?

https://w3ctag.github.io/design-principles/#events-vs-observers

p.s.
I cleaned up the code by making certain adjustments. 🤓

tihuan: Ooh thanks for sharing the link! That doc looks great ⚡Will need to read through the whole thing later too ahaha!

Yeah I think you're right that using resize event here is better, because we're not getting any element dimensions anyway, so no extra js loops saved 😆 Thanks for pointing that out!

tihuan:
One last thing 🙏 Should we use throttle over denounce in resize event? This way we can resize the page some times while the window is being resized, but not too many times to degrade performance!

.storybook/manager-head.html
masoudmanson: Thanks for checking that out 🙏🏻🥳

masoudmanson:
> One last thing 🙏 Should we use throttle over denounce in resize event? This way we can resize the page some times while the window is being resized, but not too many times to degrade performance!

Hmm, sounds like a good idea, I'll test it out 🤩🥳

Issue #356, opened 2023-02-14
Label on TooltipTableProps is nullable, but leads to non-unique react keys and console errors.


seve:
https://github.com/chanzuckerberg/sci-components/blob/b9a36626a821ee5f2e12295665d3f5035f9f521e/src/core/TooltipTable/style.ts#L14

https://github.com/chanzuckerberg/sci-components/blob/b9a36626a821ee5f2e12295665d3f5035f9f521e/src/core/TooltipTable/index.tsx#L26

```
Warning: Encountered two children with the same key, `undefined`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

PR #365, opened 2023-02-23
fix(tooltiptable): fix non-unique react keys

masoudmanson:
## Summary

**TooltipTable**
Shortcut ticket: #356

### DIFF
--- a/src/core/TooltipTable/index.tsx
+++ b/src/core/TooltipTable/index.tsx
@@ -25,7 +25,7 @@ const TooltipTableContent = (props: TooltipTableContentProps): JSX.Element => {
       {data?.map((section, index) => (
         <Section
           disabled={section.disabled}
-          key={`${section.label?.concat(index.toString())}`}
+          key={`${section.label + String(index)}`}
         >
           <SectionLabel disabled={section.disabled} label={section.label}>
             {section.label}
@@ -33,7 +33,7 @@ const TooltipTableContent = (props: TooltipTableContentProps): JSX.Element => {
           <Table size="small">
             <TableBody>
               {section.dataRows.map((row, rowIndex) => (
-                <TableRow key={row.label.concat(rowIndex.toString())}>
+                <TableRow key={row.label + String(rowIndex)}>
                   <RowLabel disabled={section.disabled} align="left">
                     {row.label}
                   </RowLabel>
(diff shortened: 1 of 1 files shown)


src/core/TooltipTable/index.tsx
```diff
@@ -22,15 +22,18 @@ const TooltipTableContent = (props: TooltipTableContentProps): JSX.Element => {
   return (
     <TableContainer {...rest}>
       {contentAlert && <Alert>{contentAlert}</Alert>}
-      {data?.map((section) => (
-        <Section disabled={section.disabled} key={`${section.label}`}>
+      {data?.map((section, index) => (
+        <Section
+          disabled={section.disabled}
+          key={`${section.label?.concat(index.toString())}`}
```
tihuan, line 28: super nit: Maybe this? Personally seems more readable 😄 
```suggestion
          key={section.label + String(index)}
```

src/core/TooltipTable/index.tsx
```diff
@@ -22,15 +22,18 @@ const TooltipTableContent = (props: TooltipTableContentProps): JSX.Element => {
...
+        >
           <SectionLabel disabled={section.disabled} label={section.label}>
             {section.label}
           </SectionLabel>
           <Table size="small">
             <TableBody>
-              {section.dataRows.map((row) => (
-                <TableRow key={row.label}>
+              {section.dataRows.map((row, rowIndex) => (
+                <TableRow key={row.label.concat(rowIndex.toString())}>
```
tihuan, line 36:
```suggestion
                <TableRow key={row.label + String(rowIndex)}>
```

tihuan (approved):
Looks v good! Thanks for the quick fix, Masoud 🙏 One tiny comment, otherwise LGTM 🚀

src/core/TooltipTable/index.tsx
tihuan: Also prefer `String(foo)` over `foo.toString()`, so when `foo` is `undefined` or `null`, it won't throw a runtime TypeError and will just render string `undefined ` and `null` instead!

masoudmanson: Woow ,Thanks for the suggestion. I just refactored it 🥇 
Much better now :D

Issue #403, opened 2023-03-30
Clicking outside of a dropdown - Update to expected behavior


cathystoli:
Problem: When you click out of a `DropdownMenu` component it applies changes
 
Slack convo: https://czi-sci.slack.com/archives/C032S43KKFV/p1679696875607709 

Figma & expected behavior: https://www.figma.com/file/35JMdsMEL252zYWthRG1ui/Dropdown-Variation---Apply-button?node-id=15%3A66531&t=N78JbQ8OlEtY4x4O-1

cathystoli:
We left some comments in the figma based off a convo in sprint planning

clarsen-czi:
As far as I understood it, the DropdownMenu variant that includes buttons is a CZ ID snowflake. Because of this, whatever functionality the team wants to assign when users click outside of the menu is up to them, but shouldn't affect how the SDS component works.

clarsen-czi:
Just chatted with Jenn and she helped remind me that we do have the variant with buttons as part of the SDS codebase, we just don't have it documented in Zeroheight and the "click away" behavior needs to be updated.

What needs to be done:

- Documentation needs to be updated to include variant with Buttons as an SDS component (not just CZ ID)
    - Including explanation of what's supposed to happen when the user clicks outside of the dropdown (specs outlined in the Figma file above)

PR #412, opened 2023-04-03
feat(dropdownmenu): adding clickAwayListenerProps to manage clickAway behavior

masoudmanson:
## Summary

**DropdownMenu**
Github issue: #403 

The latest feature added to the Dropdown component is the ability to customize its click-away behavior using the `ClickAwayListenerProps` prop. This new prop allows passing through the Mouse and Pointer event handling props to the underlying Material UI `ClickAwayListener` component. By setting the mouseEvent and touchEvent props to false, the Dropdown menu can remain open even when users click outside of it.

To use this feature, include `ClickAwayListenerProps` as part of the `DropdownMenuProps`. For example, the following code shows how to create a Dropdown with a customized click-away behavior:

```javascript
<Dropdown
  label="Click Target"
  onChange={() => {}}
  options={GITHUB_LABELS}
  DropdownMenuProps={{
    ClickAwayListenerProps: {
      mouseEvent: false,
      touchEvent: false,
    },
  }}
/>
```

With this implementation, the Dropdown menu will remain open when users click outside of it using both mouse and touch events.

The `DropdownMenu` component can also achieve the same customized click-away behavior. For instance, you can use the `ClickAwayListenerProps` prop to disable the default behavior of closing the menu when users click outside of it.

Here's an example of a `DropdownMenu` component with a customized click-away behavior:

```javascript
<DropdownMenu
  multiple={multiple}
  onChange={handleChange}
  options={options}
  search={search}
  title={title}
  ClickAwayListenerProps={{
    mouseEvent: false,
    touchEvent: false,
  }}
  {...props}
/>

```

With the above code, the DropdownMenu will remain open even when users click outside of it using both mouse and touch events, thanks to the `mouseEvent` and `touchEvent` props set to false in the `ClickAwayListenerProps` prop.

The `mouseEvent` prop of the `clickAwayListenerProps` can be set to one of the following values:

- `onClick`
- `onMouseDown`
- `onMouseUp`
- `onPointerDown`
- `onPointerUp`
- `false`

The default value for mouseEvent is `onClick`.

Similarly, the `touchEvent` prop of the `ClickAwayListenerProps` can be set to one of the following values:

- `onTouchEnd`
- `onTouchStart`
- `false`

The default value for touchEvent is `onTouchEnd`.

More information about these props can be found here: https://mui.com/base/api/click-away-listener/

## Checklist

- [x] Test Story in Storybook
- [x] Tests written


src/core/DropdownMenu/index.tsx
```diff
@@ -67,6 +68,7 @@ interface ExtraDropdownMenuProps extends StyleProps {
   PaperComponent?: typeof StyledPaper | RenderFunctionType;
   children?: JSX.Element | null;
   onClickAway: (event: MouseEvent | TouchEvent) => void;
+  clickAwayListenerProps?: Partial<ClickAwayListenerProps>;
```
tihuan, line 71: nit: Can we use capital `C` instead to signify it's a component prop please 🙏 ? Following what MUI does and other props we have! Thanks so much!

```suggestion
  ClickAwayListenerProps?: Partial<ClickAwayListenerProps>;
```

tihuan (approved):
One tiny change to prop name otherwise LGTM!

src/core/DropdownMenu/index.tsx
masoudmanson: Aah, for sure. Thanks for spotting this. 🥳

masoudmanson: It appears that there is already a ClickAwayListenerProps defined in MUI and is a reserved name. Therefore, we need to come up with a different name for our prop. Would it make sense to rename it to SdsClickAwayListenerProps instead? @tihuan

cathystoli:
Release must have checklist:
- [ ] Zeroheight minimum documentation
- [ ] QA passing

Should haves:
- [ ] Figma component for DropDownmenu
- [ ] Component recipes

PR #430, opened 2023-04-19
fix(dropdown): if value prop has changed, update value state

ehoops-cz:
Previously, passing a new array as a value prop would trigger an infinite loop with the propValue and value useEffects. Now, when the Dropdown receives a value prop, it checks to see if it is different from the current value state before updating. This allows the component to be controlled by the parent.

## Summary

**Structural Element (Base, Gene, DNA, Chromosome or Cell)** Dropdown
Github issue: none
Copy isuue description: Issue Description: Can someone please help me understand how to control the dropdown state.  I am passing the 'value' prop, so I think I should be controlling which options are checked.  However, this doesn't seem to be the case. 
For example, I am currently passing an empty list as the value.  I expect that these steps: open the dropdown, select some options, close the dropdown, open the dropdown again
would result in no options being selected because I am passing in an empty list for the value.  This isn't the case, so I'm confused about how the dropdown is actually controlled.  

Note: I currently have a type error in the compare function - it would be helpful to get some guidance here.

## Checklist

- [x] Test Story in Storybook
- [ ] Tests written


src/core/Dropdown/index.tsx
```diff
@@ -114,7 +124,11 @@ const Dropdown = <Multiple extends boolean | undefined = false>({
...
   useEffect(() => {
-    if (isControlled) {
+    if (
+      isControlled &&
+      ((!multiple && propValue !== value) ||
+        (multiple && !compareOptionArrays(propValue, value)))
+    ) {
       setValue(propValue);
     }
   }, [propValue]);
```
ehoops-cz, line 134: I'm not entirely sure whether this equality check should go here or above in the `value` useEffect.  That one may make more sense...

tihuan:
@ehoops-cz will close this for now for PR cleanups. We can reopen if still needed! Thank you!

PR #436, opened 2023-04-21
test: Update Button component Chromatic screenshots

liaprins:
## Updates Button component's Live Preview to show all styles, states, types in single view

**Updates for Chromatic tests**

## Checklist

- [x] Test Story in Storybook
- [x] Tests written
- [ ] Variables from `defaultTheme.ts` used wherever possible (Lia: not sure...)
- [ ] If updating an existing component, depreciate flag has been used where necessary (Lia: I think not applicable...)
- [ ] Chromatic build verified by @chanzuckerberg/sds-design (Lia: don't know how to do this yet, but this PR by definition will change the existing previous Chromatic screenshot for Live Preview)

Now the Button component has a single Live Preview story page that houses: 
- all the button styles (rounded, square, minimal) 
- × all the button types (primary, primary with icon, secondary, secondary with icon (except minimal style doesn't have this last one)
- × whether the button has an icon or not
- ×  whether the button is disabled or not (added May 3)
- × all their states (default, hover, active, focus-visible)

This will make it simpler for Chromatic testing: there will just need to be a single screenshot of this story page to compare every possible button permutation against in one go.

**Update (May 3):** 
I have updated the Button Live Preview to be more similar to the Tag's Live Preview (PR #445 ), which I worked on right after this Button story initially. Now the Button Live Preview story includes permutations for whether it is disabled or not, as well as labels for each of the permutations' dimensions. The way the icon is called or not is also more "DRY".

new live preview layout

### DIFF
--- a/packages/components/src/core/Button/index.stories.tsx
+++ b/packages/components/src/core/Button/index.stories.tsx
@@ -64,6 +64,8 @@ const placementStyles = {
   gridTemplateRows: "1fr",
 };
 
+const PSEUDO_STATES = ["default", "hover", "active", "focus-visible"];
+
 export const RoundedLivePreview = {
   parameters: {
     snapshot: {
@@ -72,32 +74,49 @@ export const RoundedLivePreview = {
   },
   render: (props: Args): JSX.Element => {
     return (
-      <div style={placementStyles as React.CSSProperties}>
-        <RawButton {...props} sdsStyle="rounded" sdsType="primary">
-          {text}
-        </RawButton>
-
-        <RawButton
-          {...props}
-          startIcon={<Icon sdsIcon="download" sdsSize="s" sdsType="button" />}
-          sdsStyle="rounded"
-          sdsType="primary"
-        >
-          {text}
-        </RawButton>
-        <RawButton {...props} sdsStyle="rounded" sdsType="secondary">
-          {text}
-        </RawButton>
-        <RawButton
-          {...props}
-          startIcon={<Icon sdsIcon="download" sdsSize="s" sdsType="button" />}
-          sdsStyle="rounded"
-          sdsType="secondary"
-        >
-          {text}
-        </RawButton>
-      </div>
+      <>
+        {PSEUDO_STATES.map((state) => {
+          return <ButtonRow state={state} />;
+        })}
+      </>
     );
+
+    function ButtonRow({ state = "default" }) {
+      return (
+        <div id={state} className={`pseudo-${state}`}>
+          <h5>{state}</h5>
+          <div style={placementStyles as React.CSSProperties}>
+            <RawButton {...props} sdsStyle="rounded" sdsType="primary">
+              {text}
+            </RawButton>
+
+            <RawButton
+              {...props}
+              startIcon={
+                <Icon sdsIcon="download" sdsSize="s" sdsType="button" />
+              }
+              sdsStyle="rounded"
+              sdsType="primary"
+            >
+              {text}
+            </RawButton>
+            <RawButton {...props} sdsStyle="rounded" sdsType="secondary">
+              {text}
+            </RawButton>
+            <RawButton
+              {...props}
+              startIcon={
+                <Icon sdsIcon="download" sdsSize="s" sdsType="button" />
+              }
+              sdsStyle="rounded"
+              sdsType="secondary"
+            >
+              {text}
+            </RawButton>
+          </div>
+        </div>
+      );
+    }
   },
 };
 
--- a/packages/components/src/core/Button/index.stories.tsx
+++ b/packages/components/src/core/Button/index.stories.tsx
@@ -5,8 +5,8 @@ import Icon from "../Icon";
 import RawButton from "./index";
 
 const text = "Label";
-const sdsStyles = ["rounded", "square", "minimal"];
-const sdsTypes = ["primary", "secondary"];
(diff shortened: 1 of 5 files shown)


packages/components/src/core/Button/index.stories.tsx
```diff
@@ -56,121 +62,193 @@ export const Default = {
 
 // Live Preview
 
-const placementStyles = {
-  display: "grid",
-  gridColumnGap: "10px",
-  gridRowGap: "0px",
-  gridTemplateColumns: "repeat(6, 120px)",
-  gridTemplateRows: "1fr",
+const styleLevel = {
```
liaprins, line 65: Adding new / replacing old styles to better lay out all the permutations for Button component in the UI, using CSS Grid

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -56,121 +62,193 @@ export const Default = {
...
+    }
+
+    // loop through all ICON_OPTIONS + create headers for SDS_TYPES
+    function ButtonTypeOption({ sdsStyle, type }) {
+      return (
+        <div style={displayContents}>
+          <h3 style={typeLabel}>
+            Type: <b>{type}</b>
+          </h3>
+          {type === "primary" ? (
```
liaprins, line 157: For the `minimal` style, there is no "secondary" sdsType button with an icon. So this code checks for the type and if it's primary, it loops through the two icon options (`undefined` and a download icon), but if not primary, it manuualy assigns the first icon option from the array of options (`undefined`, line 172)

liaprins:
I just noticed that the butoons that are both disabled and in focus state have the visible focus ring. Are disabled elements supposed to be able to be focused? 🤔 If not, need to determine if this is a bug in the component code, or a bug of using the [Storybook Pseudo State addon](https://storybook.js.org/addons/storybook-addon-pseudo-states)...

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -56,121 +56,97 @@ export const Default = {
...
-const minimalPlacementStyles = {
-  display: "grid",
-  gridColumnGap: "24px",
-  gridRowGap: "0px",
-  gridTemplateColumns: "repeat(3, min-content)",
-  gridTemplateRows: "1fr",
-};
-
-export const MinimalLivePreview = {
+export const LivePreview = {
```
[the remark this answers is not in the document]
tihuan: Perfect thank you!

tihuan (approved):
Great work on refactoring and cleaning up the tests, Lia 🎉 🙏 V Impressive!

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -134,7 +132,11 @@ export const LivePreview = {
     );
 
     // loop through all SDS_TYPES + create headers for SDS_STYLES
-    function ButtonStyleOption({ sdsStyle }) {
+    function ButtonStyleOption({
+      sdsStyle,
+    }: {
+      sdsStyle: (typeof SDS_STYLES)[number];
```
tihuan, line 138: add type definition - this is most the remaining errors from Github PR checks!

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -193,7 +210,7 @@ export const LivePreview = {
                 type={type}
                 icon={icon}
                 disabled={disabled}
-                key={disabled}
+                key={String(disabled)}
```
tihuan, line 213: key has to be either a `number` or `string`, so I use `String()` to convert the value to `string`!

https://www.w3schools.com/jsref/jsref_string.asp

tihuan (commented):
@liaprins PTAL this commit diff thank you!

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -56,121 +62,210 @@ export const Default = {
...
+      sdsStyle: (typeof SDS_STYLES)[number];
+      type: (typeof SDS_TYPES)[number];
+    }) {
+      return (
+        <div style={displayContents}>
+          <h4 style={typeLabel}>
+            Type: <b>{type}</b>
+          </h4>
+          {/* Minimal Secondary doesn't have icon button option */}
+          {sdsStyle === "minimal" && type === "secondary" ? (
```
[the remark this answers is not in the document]
tihuan: @liaprins this is the only change that affects the screenshot. I added secondary icon button examples for `rounded` and `square`, since your comment said only `minimal` style doesn't have secondary icon button. PTAL thank you!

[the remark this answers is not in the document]
liaprins: Ah good catch, thank you!

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -66,8 +66,6 @@ const styleLevel = {
   columnGap: "20px",
   display: "inline-grid",
   fontFamily: "sans-serif",
-  gridColumnTemplate:
```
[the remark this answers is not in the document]
liaprins: Oh okay, yeah I must have transposed the name,but thanks for catching and fixing!

packages/components/src/core/Button/index.stories.tsx
liaprins: Oh, perfect, thanks!!

PR #440, opened 2023-04-25
chore: Update ComplexFilter 'label' type from string to ReactNode

ovalenzuela19:
## Summary
Changes ComplexFilter 'label' type from `string` to `ReactNode`. The Dropdown has label defined as ReactNode so I expected the ComplexFilter to also be a ReactNode -- https://github.com/chanzuckerberg/sci-components/blob/main/packages/components/src/core/Dropdown/index.tsx#L46


ehoops-cz (approved):
lgtm

PR #445, opened 2023-05-01
test: Tag component's "Live Preview" story page for Chromatic tests

liaprins:
## Summary

**Tag component**
Github issue: #469 

### **PR context:**
This PR updates the output of the Tag component’s Live Preview story page in Storybook, such that it displays all permutations of the Tag component across the following dimensions:
- sdsStyle (rounded, square) × 
- color (primary, info, success, warning, error, gray, beta, custom ([“#A52A2A", "#ffa07a", "#ffe74a”])) × 
- pseudo state (default, hover, active, focus-visible) × 
- able to be hovered or not (binary) × 
- sdsType (primary, secondary) × 
- whether or not it has an icon (binary)

Having all of these possible permutations in one screen allows Chromatic to take a single screenshot of this one page to determine whether any of the Tag’s dimensions changed unintentionally.

I organized the visual layout or the page to make it as easy as possible to compare across tag permutations that would be relevant to compare across (e.g. I kept each color together, but per pseudo state, put iterations of primary/secondary type, hoverable/not hoverable, and  icon/no icon near each other, since these would likely be what we should be comparing to each other when checking the visual output).

Due to their being 256 distinct tag permutations on this Live Preview page, the viewer will need to scroll to see them all. But Chromatic should be able to still take a screenshot of everything (this is my understanding at least). If it takes screenshots that reveal unintended changes in the future, it may be difficult to see the change since there are so many iterations in the screenshot, but having the layout organized to be able to scroll through will help with examining (may need to zoom into or download the Chromatic screenshot to see it adequately, but this should still be simpler than having many, many story pages to accommodate each possible permutation / story).

Here is a screenshot I took (not Chromatic) of the Live Preview story page, I zoomed out my browser as far as possible, but it is still not fitting all the `color` sections:
zoomed-out-screenshot

Here is a GIF scrolling through all the Tag permutations:
scrolling

### **Issue with the `hover` dimension:**
To generate the pseudo state iterations of the Tags, we use a Storybook add-on that “forces” a component to look how it would for a given pseudo state, but it uses CSS. So it was not working for the “Hoverable: false” iterations for any of the pseudo-state sections: these Tags should have looked just like the `default` pseudo state regardless of their actual pseudo state (e.g. hover, active…) since they cannot be hovered upon/interacted with by definition (`hover={false}`). However, the way the pseudo state addon renders seems to interfere with this, because all the “Hoverable: false” iterations actually did NOT look like the default state (except the one that was the default state), instead they looked like the “Hoverable: true” iterations of the same pseudo-state. 

So, I am currently forcing all “Hoverable: false” Tag iterations to have their correct rendering (to look like the `default` pseudo state) by actually having the the `default` pseudo state with this line:  `className={hover ? `pseudo-${state}` : `pseudo-default`}`
(line ~331)

This is not really useful for truly testing whether a Tag with `hover={false}` would really look the way it should when hovered upon (not change at all, maintain the look of `default` pseudo state), though, because in this Storybook page it is being overridden to achieve that visual look. But I left it for now in case we could figure out how to fix it in future. Let me know if you know how we could, or if we should just remove the “able to be hovered or not” dimension entirely from the story.

There are a few other things I wasn’t 100% sure if I was doing right in terms of SDS code conventions, but I will add those as comments to this PR next!

### Panel options
I was thinking it would be a good idea to disable or remove the Storybook panel or its options from the UI, but I could not figure out how yet...

## Checklist

- [x] Test Story in Storybook
- [x] Tests written
- [ ] Variables from `defaultTheme.ts` used wherever possible (Lia: Not sure)
- [ ] If updating an existing component, depreciate flag has been used where necessary (Lia: I don't think this is necessary since it's just updating tests)
- [ ] Chromatic build verified by @chanzuckerberg/sds-design (Lia: This shoudl change the Chromatic build by definition)


packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -102,102 +102,241 @@ export const Default = {
...
+        </div>
+      );
+    }
+
+    // loop through all PSEUDO_STATES + create headers for COLORS
+    function TagHoverableOption({ sdsStyle, color }) {
+      return (
+        <div style={displayContents}>
+          <h2 style={colorLabel}>
+            Color: <b>{color[0].length === 1 ? color : "custom"}</b>
```
liaprins, line 234: This creates the text heading for each color. There was an issue with the spacing of the grid when I used the `{color}` var exclusively as the content, because the custom color, which is defined in the `COLOR` const as a sub-array of actual hex values (but the other colors are named vars) "prints out" to the UI as the three hex codes all in a row with no spaces (#A52A2A#ffa07a#ffe74a). It will not break when I tried to the CSS `hyphen: auto` so it was making a very long non-breaking word which resulted in spacing out the distance between the tags more than intended. Also, it is more useful to label this color as "Custom" rather than the hexcodes, so in this line I tried to query whether the `color` var per loop was an array but it was not working... the only thing I could figure out was to get the length of the `color` item and if it is just a length of one then it's just one word (so a predefined SDS color) rather than the custom sub-array. So point being, this line is checking whether the `color` is the custom color or not, and if so it labels it as "Custom" in the UI.

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -102,102 +102,241 @@ export const Default = {
...
+    }
+
+    // loop through all HOVERABLE_OPTIONS + create headers for PSEUDO_STATES
+    function TagColor({ sdsStyle, color, state }) {
+      return (
+        <div style={displayContents}>
+          <h3 style={stateLabel}>
+            Pseudo state: <b>{state}</b>
+          </h3>
+          {HOVERABLE_OPTIONS.map((hover, hoverIndex) => {
```
liaprins, line 257: In order to properly stack the the hoverable iterations of the Tag component ("Hoverable: yes" and "Hoverable: no") within the CSS grid for easier visual comparison, the index of the items in this `.map()` loop needs to be passed all the way down to the last loop. If this loop's index is zero, the sdsType and the Icon headings are rendered, but not otherwise (see lines 308 and 317 below). This allows multiple rows of tags (just two in this case) to be stacked with just one sdsType and Icon headings as needed, applying to the columns created by the Tag components' placements within the CSS grid.

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -102,102 +102,241 @@ export const Default = {
...
+                  <h6 style={iconLabel}>
+                    Icon: <b>{iconText}</b>
+                  </h6>
+                )}
+                <RawTag
+                  {...props}
+                  data-testid="tag"
+                  sdsStyle={sdsStyle}
+                  color={color}
+                  className={hover ? `pseudo-${state}` : `pseudo-default`}
```
liaprins, line 327: Here's the issue I mentioned in the PR description about the `hover` prop being overridden by the pseudo state addon, so I temporarily overrode it myself such that the tags that have `hover={false}` would always render with the default pseudo state...

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -102,102 +102,241 @@ export const Default = {
...
+const iconLevel2 = {
+  gridArea: "2 / 3 / 2 / 3",
+};
+
+const styleLabel = {
+  fontWeight: "normal",
+  gridColumn: "2 / 4",
+  margin: 0,
+};
+const colorLabel = {
```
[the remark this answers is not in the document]
tihuan: Yes! I think you can use object spreading to achieve what you want 👍 

E.g.,

```ts
const commonStyle = { 
  borderStyle: "solid none none none",
  fontWeight: "normal",
}

const typeLabel = {
  ...commonStyle,
  borderWidth: "2px",
}
```

tihuan:
Good news and bad news 😆 

Good news is that @liaprins 's expanded tests and Edu team's a11y test tool caught color contrast bug for us!

The bad news is that we need to fix the color contrast bug lol

It seems like many tag color + text color combos have color contrast issues, so not sure if it is something we can fix in this PR quickly or requires more time to resolve. If so, maybe we can comment out the test and get the PR merged and then after fixing the color contrast issue, reenabling the test!

Screenshot 2023-05-10 at 7 42 18 AM

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -102,102 +111,245 @@ export const Default = {
...
-          sdsType="secondary"
-        />
+    return (
+      <div style={penultimateLevel}>
+        {PSEUDO_STATES.map((state) => {
+          return (
+            <div style={bottomLevel}>
+              {(hover === true || (hover === false && state === "default")) && (
+                <>
+                  <h6 style={bottomLabel}>
```
liaprins, line 308: Removed irrelevant `hover` iterations: when combined with all pseudo-states except default, `hover=false` is impossible

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -12,8 +12,8 @@ const Tag = (props: Args): JSX.Element => {
 };
 
 const CUSTOM_COLOR_TUPLES = {
-  labelAndBack: ["#A52A2A", "#ffa07a"],
-  labelAndBackAndIcon: ["#A52A2A", "#ffa07a", "#ffe74a"],
+  labelAndBack: ["#000000", "#C65FA7"],
```
liaprins, line 15: Added new custom colors that will pass the a11y tests

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -39,6 +39,8 @@ const availableIconOptions = [
 const SDS_STYLES: ExtraTagProps["sdsStyle"][] = ["rounded", "square"];
 const COLORS = availableColorOptions.slice(0, 9) as ExtraTagProps["tagColor"][];
 COLORS.splice(7, 1);
+const GRAY_PRIMARY_COLORS = ["gray"];
```
liaprins, line 42: define separate color-related vars for the 2 x new screenshot-related stories

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -111,6 +113,106 @@ export const Default = {
 
 // Live Preview
 
+const livePreviewStyles = {
```
liaprins: Restore original LivePreview code, to prevent Zeroheight from breaking

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -269,6 +378,7 @@ function LivePreviewDemo(props: Args): JSX.Element {
         {HOVERABLE_OPTIONS.map((hover) => {
           return (
             <TagState
+              {...props}
```
liaprins, line 381: I think this is now needed here so that `...props` are passed to the next loop and it can exist externally and be used by multiple stories (at least it wasn't working at all until I added this line...)

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -282,57 +392,516 @@ function LivePreviewDemo(props: Args): JSX.Element {
     );
   }
 
-  // loop through all PSEUDO_STATES + create headers for HOVERABLE_OPTIONS, PSEUDO_STATES
-  function TagState({
+  // // loop through all PSEUDO_STATES + create headers for HOVERABLE_OPTIONS, PSEUDO_STATES
```
liaprins, line 395: I left in the commented-out parts of the 3 x screenshot-related stories for context: this is the loop that is now externalized and reused by all three, so this code doesn't need to exist redundantly for each (But I left it commented out for context for now)

tihuan (approved):
Fantastic work!! Love the description for the separated out stories too, Lia! That makes things super clear ✨ 

LGTM!! Let's merge it 🔥

PR #451, opened 2023-05-05
feat(dropdown): new tests for the dropdown component

masoudmanson:
- [x] Adding new test cases for the Dropdown component
- [x] Fixing closeOnBlur bug: Dropdown closes when button is clicked with closeOnBlue set to true


packages/components/src/core/Dropdown/index.test.tsx
```diff
@@ -40,4 +61,154 @@ describe("<Dropdown />", () => {
...
+    const args = {
+      ...Test.args,
+      closeOnBlur: true,
+    };
+
+    render(<Test {...args} />);
+
+    // Open the dropdown
+    const DropdownElement = screen.getByTestId("dropdown");
+    userEvent.click(DropdownElement);
```
liaprins, line 129: Doe sit actually need to use `userEvent` here, or would `fireEvent` work now that it has `wait(500)` below?

packages/components/src/core/Dropdown/index.test.tsx
```diff
@@ -40,4 +61,154 @@ describe("<Dropdown />", () => {
...
+
+    // Awaiting to find the STATUS_CANT_FIND_TEXT on the screen to
+    // make sure that the dropdown is opened
+    await screen.findByText(STATUS_CANT_FIND_TEXT);
+
+    // Click on the button again to close it
+    userEvent.click(DropdownElement);
```
liaprins, line 136: Same question about `userEvent` here as above...?

liaprins (approved):
I added a couple of comments, but it looks good to me!

packages/components/src/core/Dropdown/index.test.tsx
masoudmanson: Actually, it doesn't work with fireEvent, and we kinda have to use userEvent in this case!

liaprins: Got it!

Issue #453, opened 2023-05-10
Chromatic test for ButtonIcon


liaprins:
### Alignment
- [x] Align with design (and product, other engineers, if necessary) on the visual traits to include as permutation dimensions for the Chromatic test
- Find the component in this doc as a starting point: [Chromatic tests: Which component dimensions to consider looping through](https://docs.google.com/document/d/1i40YV1rX61dNzGsqbhqJAD8PL1puCecX7rnHgD_FawE/edit#)
- Permutation dimensions may include the component’s props, but it may not always be necessary to include every prop
- If the component can be interacted with, or if part of it can (e.g. an “x” button or similar within it), it should include pseudo state as a permutation dimension

- [x] Once the permutation dimensions are aligned on, list them in this issue as checkboxes to track them

### Permutation dimensions to include
- [x] **sdsType**: Primary, Secondary, Tertiary
- [x] **sdsSize**: Large, Medium, Small
- [x] **on**: false, true
- [x] **disabled**: false, true
- [x] **pseudo-state**: default, hover, active, focus

#### Exceptions, notes
- [x] sdsType Primary and Secondary do _not_ have sdsSize Medium
- [x] Secondary x Large and all Tertiary sizes do _not_ have the "on" dimension
- [x] No on/off iteration for primary x small ([UPDATE 05/22/23 PER PR COMMENT BY CONNOR](https://github.com/chanzuckerberg/sci-components/pull/462#issuecomment-1557676343))
- [x] Don't use the same icon throughout; instead, vary the icon to match known usecases in product according to this screenshot:
![Screenshot 2023-05-10 at 11 35 50](https://github.com/chanzuckerberg/sci-components/assets/17435353/b4b7c61a-b00f-40b7-9639-491a54f49cf8)
- [x] Move `disabled=true` (for default pseudo state only) to be visually part of the pseudo-state row, in terms of layout in the UI
- Disabled is not yet needed in product for Tertiary of any size, however, I will still include it in the Chromatic test (and designs do exist for it)
- ButtonIcon uses the default browser focus state

### Tips for writing Chromatic tests
- Write the tests in index.stories.tsx within the component’s folder
- Name the story page LivePreview `export const LivePreview ... `
- Use or create an array per permutation dimension and loop through each to iteratively create all relevant combinations of the component
- Put pseudo-state at the lowest level
- See Button's Live Preview story for an example
[...]

PR #462, opened 2023-05-12
test(ButtonIcon): ButtonIcon screenshot test

liaprins:
## Summary

**ButtonIcon**
Github issue: #453 

## Notes about this PR
This PR updates the ButtonIcon's Live Preview page so that every viable permutation of ButtonIcon component is included, so we can automatically catch (via Chromatic) any potential unintended changes to any combinations of a component’s visually relevant props.

Screenshot 2023-05-23 at 9 08 43 AM

- Note that not every dimension-combination is relevant for ButtonIcon (e.g. `sdsType`'s primary and secondary have no `sdsSize` of medium, and not all have the `on` prop); only the relevant permutations are included and labeled.
- Also note that I included the `disabled` iteration in the same row, visually, as the pseudo-states. Disabled elements cannot be interacted with so cannot ever have any pseudo-state other than default, so it does not make sense to loop the `disabled = true` status through every pseudo-state iteration, only through the default state. Furthermore, the Storybook addon we use to simulate pseudo states does not properly render the states when `disabled=true` (it shows them as having hover, active, and focus statues visually which is not actually possible). So I removed every permutation of the ButtonIcon for `disabled=true`, except for the case where pseudo-state is default. Then I put this instance in the same row in the UI as the other pseudo-state iterations of ButtonIcon; this matches how components are shown in Zeroheight as well.

## GitHub issue
- [x] Align with design (and product, other engineers, if necessary) on the visual traits to include as permutation dimensions for the Chromatic test

- [x] sdsType: Primary, Secondary, Tertiary
- [x] sdsSize: Large, Medium, Small
- [x] on: false, true
- [x] disabled: false, true
- [x] pseudo-state: default, hover, active, focus

#### Exceptions, notes
- [x] sdsType Primary and Secondary do not have sdsSize Medium
- [x] Secondary x Large and all Tertiary sizes do not have the "on" dimension
- [x] No on/off iteration for primary x small (https://github.com/chanzuckerberg/sci-components/pull/462#issuecomment-1557676343)
- [x] Don't use the same icon throughout; instead, vary the icon to match known use cases in product according to icons used in Zeroheight 
- [x] Move `disabled=true` (for default pseudo state only) to be visually part of the pseudo-state row, in terms of layout in the UI
- Disabled is not yet needed in product for Tertiary of any size, however, I will still include it in the Chromatic test (and designs do exist for it)
- ButtonIcon uses the default browser focus state

## Checklist
- [x] Default Story in Storybook (liaprins: Default story exists, but this PR only includes updates to the Live Preview story)
- [x] Test Story in Storybook (liaprins: Test story exists, but this PR only includes updates to the Live Preview story)
- [x] Tests written (liaprins: only wrote the test for Live Preview)
- [x] Variables from `defaultTheme.ts` used wherever possible (liaprins: I think so?...)
- [ ] If updating an existing component, depreciate flag has been used where necessary (liaprins: N/A)
- [ ] Chromatic build verified by @chanzuckerberg/sds-design (liaprins: not sure; by definition this PR intends to update the Chromatic test)


liaprins:
@clarsen-czi Just tagging you in case you want to verify if the permutations rendered correctly (see screenshot); I checked against the designs in Zeroheight as I worked, but feel free to have a look if you want to make sure everything is in proper working order in terms of how the code renders.

liaprins:
Just realized this is failing one of the tests for a11y, but I don't know why; all the H6 headers are adjacent to each other in terms of nesting 🤔 ... if anyone has any insights let me know!

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -1,5 +1,6 @@
 import { Args, Meta } from "@storybook/react";
 import React from "react";
+import { defaultAppTheme } from "../styles";
```
liaprins, line 3: Needed for original `LivePreview` test

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -93,245 +94,362 @@ export const Default = {
...
-  ...midLabel,
-  alignSelf: "end",
-  borderWidth: "1px",
-  fontWeight: "normal",
-  margin: "0 0 5px 0",
-};
-const bottomLabel: React.CSSProperties = {
-  ...fontWeightNormal,
-  margin: "10px 0",
+const LivePreviewDemo = (): JSX.Element => {
```
liaprins, line 97: Restores the original `LivePreview` test so Zeroheight won't break

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -93,245 +94,362 @@ export const Default = {
...
+    borderWidth: "1px",
+    fontWeight: "normal",
+    margin: "0 0 5px 0",
+  };
+  const bottomLabel: React.CSSProperties = {
+    ...fontWeightNormal,
+    margin: "10px 0",
+  };
 
-  render: (props: Args): JSX.Element => {
```
liaprins: This goes down lower in the code now, becuase I switched it to first build a "...Demo" component which is then called by the main component lower down (line 445).

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -195,6 +210,250 @@ export const LivePreview = {
...
+    );
+  }
+
+  // }, // close render
+}; // close ScreenshotTest
+
+export const ScreenshotTest = {
+  parameters: {
+    controls: {
+      exclude: ["disabled", "on", "sdsIcon", "sdsSize", "sdsType"],
```
liaprins, line 448: Disables the control panel in the UI since it is not needed on the Chromatic screenshot story page

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -195,6 +210,246 @@ export const LivePreview = {
...
+    display: "contents",
+  };
+  const penultimateLevel: React.CSSProperties = {
+    display: "contents",
+  };
+  const bottomLevel: React.CSSProperties = {
+    marginBottom: 10,
+  };
+  const topLabel: React.CSSProperties = {
+    fontSize: "2em",
```
liaprins, line 232: Assigns font-size to the top-level label, since all the labels use `<p>` tags now instead of `<h1>`, etc, so that labels can still be conditional and pass the a11y test (which requires headings to only increase by one level, not possible in the case of conditional heading elements)

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -195,6 +210,246 @@ export const LivePreview = {
...
+        return <ButtonIconTypeOption sdsType={sdsType} key={sdsType} />;
+      })}
+    </>
+  );
+
+  // loop through all SDS_SIZES + create headers for SDS_TYPES
+  function ButtonIconTypeOption({ sdsType }: { sdsType: SDSTypes[number] }) {
+    return (
+      <div style={topLevel}>
+        <p style={topLabel}>
```
liaprins, line 274: All labels use `<p>` instead of `<h1>` now so that conditionally used labels don't break the a11y test, since headings can only increase by one to be a11y compliant

tihuan: SGTM! Thanks for the solution, Lia!

packages/components/src/core/ButtonIcon/index.stories.tsx
tihuan, line 232: maybe move the constants closer to their first use and rename them with ALL_CAPS. There seems to be a few that can be declared inside their corresponding component functions too!

E.g.,

```tsx
function ButtonIconSizeOption() {
  const SECOND_LABEL = {
    ...midLabel,
    borderWidth: "2px",
    fontSize: "1.17em",
    margin: "20px 0",
  }

  return (
    ...
     <p style={SECOND_LABEL}>
  )
}
```

Also, to improve readability, if the variables are declared at their corresponding component level, perhaps we can even rename them?

For example, currently we have `secondLabel`, which doesn't convey much meaning. Now that the variable is declared inside `<ButtonIconSizeOption />`, perhaps we can rename it to `LABEL_STYLE`?

```tsx
function ButtonIconSizeOption() {
  const LABEL_STYLE = {
    ...midLabel,
    borderWidth: "2px",
    fontSize: "1.17em",
    margin: "20px 0",
  }

  return (
    ...
     <p style={LABEL_STYLE}>
  )
}
```

tihuan (commented):
Awesome work, Lia! Just one nit comment on variable naming and where to declare them. PTAL thank you!

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -195,6 +210,246 @@ export const LivePreview = {
...
+    return (
+      <div style={DISPLAY_CONTENTS}>
+        <p style={LABEL_STYLE}>
+          Size: <b>{sdsSize}</b>
+        </p>
+        {ON_OPTIONS.map((on) => {
+          return (
+            // for the combinations with `on` as a prop, loop through all values for `on`(true, false)
+            // for the combinations without `on` as a prop, loop through only once
+            (ON_LABEL_NEEDED ||
```
liaprins, line 301: Renamed this var to be in the ALL_CAPS style

clarsen-czi:
> @clarsen-czi Just tagging you in case you want to verify if the permutations rendered correctly (see screenshot); I checked against the designs in Zeroheight as I worked, but feel free to have a look if you want to make sure everything is in proper working order in terms of how the code renders.

So sorry for just now seeing this! Overall everything looks good, but there are a few tweaks to the states to make.

- Primary Small
    - Doesn't have an on/off state
    - Hover and active states are the same (primary/600)
- Secondary Large
    - Active state should match hover state (primary/400)
- Secondary Small
    - Active state, whether on or off, should be primary/400

Thanks @liaprins !

liaprins:
Thanks for the detailed check, @clarsen-czi ! I think this test is actually revealing some bugs in the underlying component code itself, in terms of how some of the active and / or hover states render (which is actually a good outcome of building these tests and having you check them out!). In this PR I am just inputting the various props to iterate through the relevant permutations, and then the component is rendering them. The test should automatically update if those component-code’s bugs are fixed.

I will bring this up at standup, and can then create GH issues to track the bugs (assuming I am correct about them being underlying component-code bugs).

What is in the scope of the test for me to fix is removing the on/off trait from primary x small (the first thing you noted in the PR comment), so I will fix that next!

tihuan (approved):
LGTM! Thanks so much for the super well documented PR, @liaprins !

Did you get a chance to work with Connor to create ticket(s) for the bugs too 🙏 ?

Appreciate it!

liaprins:
> Did you get a chance to work with Connor to create ticket(s) for the bugs too 🙏 ?

@tihuan Yeah, and I figured out how to fix the bugs, too! Here's the GH issue for them:
https://github.com/chanzuckerberg/sci-components/issues/477

And here's the PR I made that fixes the issue (Masoud already approved it):
https://github.com/chanzuckerberg/sci-components/pull/478

tihuan:
> > Did you get a chance to work with Connor to create ticket(s) for the bugs too 🙏 ?
> 
> @tihuan Yeah, and I figured out how to fix the bugs, too! Here's the GH issue for them: #477
> 
> And here's the PR I made that fixes the issue (Masoud already approved it): #478

Fantastic! You're so on top of it 🔥 Thanks so much!

PR #463, opened 2023-05-12
test(button): update Button Chromatic test's disabled iterations

liaprins:
## Summary
**Button**
Github issue: N/A

I updated the Live Preview story for Button so that the disabled iteration is in the same row, visually, as the pseudo-states. Disabled elements cannot be interacted with so cannot ever have any pseudo-state other than default, so it does not make sense to loop the `disabled = true` status through every pseudo-state iteration, only through the default state. Furthermore, the Storybook addon we use to simulate pseudo states does not properly render the states when `disabled=true` (it shows them as having hover, active, and focus statues visually which is not actually possible). So I removed every permutation of the Button for `disabled=true`, except for the case where pseudo-state is default. Then I put this instance in the same row in the UI as the other pseudo-state iterations of ButtonIcon; this matches how components are shown in Zeroheight as well.

This update makes more sense because it's not actually possible to have disabled x hover, disabled x active, or disabled x focus states, however these had been rendered in the Live Preview story, which was misleading.

More about the disabled attribute for context:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
- https://johnsweetaccessibility.com/2020/07/can-i-focus-disabled-controls/

Other updates:
- Changed the names of the style `const`s to be more generic, for potential reuse and / or adoption of reusable styles for across multiple Live Preview stories
- Removed control panel options since changing them didn't actually do anything, and this story page already shows every possible relevant combination of options anyway

**New layout, with single disabled component per iteration:**
Screenshot 2023-05-12 at 3 42 04 PM

**Old layout, with false focus state shown upon disabled components:**
old-layout

## Checklist

- [x] Test Story in Storybook
- [x] Tests written

### DIFF
--- a/packages/components/src/core/Button/index.stories.tsx
+++ b/packages/components/src/core/Button/index.stories.tsx
@@ -199,56 +199,204 @@ export const MinimalLivePreview = {
 
 // Screenshot Test
 
-const topLevel: React.CSSProperties = {
-  columnGap: "20px",
-  display: "inline-grid",
-  fontFamily: "sans-serif",
-  marginRight: "50px",
-};
-const displayContents: React.CSSProperties = {
-  display: "contents",
-};
-const penultimateLevel: React.CSSProperties = {
-  display: "contents",
-};
-const bottomLevel: React.CSSProperties = {
-  marginBottom: 10,
-};
-const fontWeightNormal: React.CSSProperties = {
-  fontWeight: "normal",
-};
-const topLabel: React.CSSProperties = {
-  ...fontWeightNormal,
-  fontSize: "2em",
-  gridColumn: "1 / 6",
-  marginBottom: 0,
-};
-const midLabel: React.CSSProperties = {
-  ...fontWeightNormal,
-  borderStyle: "solid none none none",
-  gridColumn: "1 / 6",
-  justifySelf: "stretch",
-  paddingTop: 10,
-};
-const secondLabel: React.CSSProperties = {
-  ...midLabel,
-  borderWidth: "2px",
-  fontSize: "1.17em",
-  margin: "20px 0",
-};
-const thirdLabel: React.CSSProperties = {
-  ...midLabel,
-  alignSelf: "end",
-  borderWidth: "1px",
-  fontWeight: "normal",
-  margin: "0 0 5px 0",
-};
-const bottomLabel: React.CSSProperties = {
-  ...fontWeightNormal,
-  margin: "10px 0",
+const ScreenshotTestDemo = (props: Args): JSX.Element => {
+  const topLevel: React.CSSProperties = {
+    columnGap: "20px",
+    display: "inline-grid",
+    fontFamily: "sans-serif",
+    marginRight: "50px",
+  };
+  const displayContents: React.CSSProperties = {
+    display: "contents",
+  };
+  const penultimateLevel: React.CSSProperties = {
+    display: "contents",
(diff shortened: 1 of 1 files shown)


packages/components/src/core/Button/index.stories.tsx
```diff
@@ -225,41 +223,40 @@ export const LivePreview = {
...
             return (
-              <div style={stateLevel}>
-                {disabledIndex % 2 ? (
-                  false
-                ) : (
-                  <h6 style={stateLabel}>
-                    State: <b>{state}</b>
-                  </h6>
+              <div style={bottomLevel}>
+                {(disabled === false ||
```
tihuan, line 237: @liaprins maybe add a comment to explain the purpose of the condition!

liaprins: Removed irrelevant `disabled` iterations: when combined with all pseudo-states except default, `hover=false` is impossible

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -60,63 +60,199 @@ export const Default = {
   },
 };
 
-// Live Preview
+// Rounded Live Preview
 
-const styleLevel = {
+const placementStyles: React.CSSProperties = {
```
liaprins, line 65: Added back in the original Live Preview tests, but converted them to first create a "...Demo" component which is subsequently called by the main LivePreview components

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -60,63 +60,199 @@ export const Default = {
...
+// Screenshot Test
+
+const topLevel: React.CSSProperties = {
   columnGap: "20px",
   display: "inline-grid",
   fontFamily: "sans-serif",
   marginRight: "50px",
 };
-const displayContents = {
+const displayContents: React.CSSProperties = {
```
liaprins: Assigned type inline for style `const`s

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -225,41 +360,41 @@ export const LivePreview = {
...
             return (
-              <div style={stateLevel}>
-                {disabledIndex % 2 ? (
-                  false
-                ) : (
-                  <h6 style={stateLabel}>
-                    State: <b>{state}</b>
-                  </h6>
+              <div style={bottomLevel}>
+                {/* removes irrelevant disabled iterations: when combined with all pseudo-states except default, `disabled=false` is impossible */}
```
liaprins: Added comment about condition as suggested by @tihuan !

packages/components/src/core/Button/index.stories.tsx
```diff
@@ -60,213 +60,352 @@ export const Default = {
...
     },
   },
-  render: (props: Args): JSX.Element => {
-    // loop through all SDS_STYLES
+  render: (args: Args) => <MinimalLivePreviewDemo {...args} />,
+};
+
+// Screenshot Test
+
+const ScreenshotTestDemo = (props: Args): JSX.Element => {
```
liaprins, line 202: Create a "...Demo" component which is then called by the main test component (line 396)

tihuan (approved):
Amazing work!! Thanks so much for improving the tests here, @liaprins 👏 🏆 !!

liaprins:
GH issue for Button Chromatic test: #468

Issue #426, opened 2023-04-08
Implement: Jump-to Navigation


clarsen-czi:
Spec sheet for implementation [here](https://www.figma.com/file/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=10902%3A107306&t=5FJpsR1HkbDz5y1U-1).

---

- Design ticket: #378 
- Documentation ticket: #380

clarsen-czi:
@cathystoli I marked this as a P1 because Single Cell will be needing this component soon, so I think ideally let's work on it next sprint?

cathystoli:
@clarsen-czi that sounds right, just added to up next

masoudmanson:
@cathystoli Just got started on this ticket.

PR #464, opened 2023-05-16
feat(navigationjumpto): implementing NavigationJumpTo component

masoudmanson:
## Summary

**NavigationJumpTo**
Github issue: #426 

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/src/common/helpers/scrollStop.ts
+++ b/packages/components/src/common/helpers/scrollStop.ts
@@ -1,0 +1,58 @@
+import { useEffect, useRef } from "react";
+
+/**
+ * Creates a scroll stop listener with a given callback and timeout.
+ *
+ * @param callback - The callback function to be executed when scrolling stops.
+ * @param timeout - The duration (in milliseconds) after which the callback is triggered when scrolling stops. Default is 50ms.
+ * @returns A function to destroy the scroll stop listener and clean up the event listener.
+ */
+const createScrollStopListener = (callback: () => void, timeout = 50) => {
+  let removed = false;
+  let handle: NodeJS.Timeout | null = null;
+
+  const onScroll = () => {
+    if (handle) {
+      clearTimeout(handle);
+    }
+    handle = setTimeout(callback, timeout);
+  };
+
+  window.addEventListener("scroll", onScroll);
+
+  return () => {
+    if (removed) {
+      return;
+    }
+
+    removed = true;
+
+    if (handle) {
+      clearTimeout(handle);
+    }
+
+    window.removeEventListener("scroll", onScroll);
+  };
+};
+
+/**
+ * Custom hook that listens for scroll stop events and triggers a callback function.
+ *
+ * @param callback - The callback function to be executed when scrolling stops.
+ * @param timeout - The duration (in milliseconds) after which the callback is triggered when scrolling stops. Default is 50ms.
+ */
+const useScrollStopListener = (callback: () => void, timeout = 50) => {
+  const callbackRef = useRef<() => void>();
+  callbackRef.current = callback;
+
+  useEffect(() => {
+    const destroyListener = createScrollStopListener(() => {
+      if (callbackRef.current) {
+        callbackRef.current();
+      }
+    }, timeout);
+    return () => destroyListener();
+  }, []);
+};
+
+export default useScrollStopListener;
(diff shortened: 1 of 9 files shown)


clarsen-czi (approved):
LGTM!! Thanks @masoudmanson !

packages/components/src/core/NavigationJumpTo/useIntersection.ts
```diff
@@ -0,0 +1,81 @@
...
+     * @param entries - IntersectionObserverEntry array containing information about each observed element.
+     */
+    const observerCallback = (entries: IntersectionObserverEntry[]) => {
+      entries.forEach((entry, index) => {
+        const name = entry.target.getAttribute("id") || index;
+
+        if (entry.isIntersecting) {
+          // Element is in view
+          setElements((prev) => {
+            return {
```
tihuan, line 44: should we add a guard clause to not set element if the item is already in view?

E.g.,

```ts
setElements((prev) => {
  if (prev[name]?.isInView) return prev;

  return {
    ...prev,
    [name]: {
      el: entry,
      isInView: true,
    },
  }
})

```

packages/components/src/core/NavigationJumpTo/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      items[newValue]?.elementRef?.current?.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+      });
+    };
+
+    const sectionIsInView = useInView(items);
+
+    useEffect(() => {
+      const sectionInView = Object.entries(sectionIsInView).findIndex(
```
tihuan, line 52: maybe just `Object.values()`?

packages/components/src/core/NavigationJumpTo/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  items: Array<{
+    title: string;
+    elementRef: React.MutableRefObject<HTMLElement | null>;
+  }>;
+}
+
+const NavigationJumpTo = forwardRef<HTMLButtonElement, NavigationJumpToProps>(
+  (props, ref): JSX.Element | null => {
+    const { items, indicatorColor, ...rest } = props;
+    const [value, setValue] = React.useState(0);
```
tihuan, line 24: maybe rename the variable to `firstTabIndexInview` for readability?

tihuan (commented):
Looks super good! Thanks so much, Masoud 😄 👏 ! Just some easy small comments otherwise LGTM!

packages/components/src/common/helpers/scrollStop.ts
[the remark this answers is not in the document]
masoudmanson: We have a use case for it in the latest commit:

```
// Set navItemClicked to false to re-enable the option
// to update the tab value based on scroll.
useScrollStopListener(() => {
    setNavItemClicked(false);
});
```

packages/components/src/core/NavigationJumpTo/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <StyledTabs
+        ref={ref}
+        orientation="vertical"
+        variant="fullWidth"
+        value={value}
+        onChange={handleChange}
+        aria-label="navigation-jump-to"
+        sdsIndicatorColor={indicatorColor}
+        {...rest}
+        indicatorColor="primary"
```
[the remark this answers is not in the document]
masoudmanson: Oh, correct, that is not necessary! :D

packages/components/src/core/NavigationJumpTo/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      >
+        {items.map(({ title, elementRef }, index) => (
+          <StyledTab
+            disableRipple
```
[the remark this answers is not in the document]
masoudmanson: Thank you for catching that! I have just added it to the defaultTheme. 🥇

packages/components/src/core/NavigationJumpTo/useIntersection.ts
masoudmanson: You raise a valid point. However, the IntersectionObserver callback is triggered only when there is a change in the elements intersecting with the view. As a result, there is already an inherent guard mechanism in place, similar to what you suggested. Therefore, I believe it is not necessary to explicitly include an additional guard clause in this case. :D

packages/components/src/core/NavigationJumpTo/index.tsx
masoudmanson: Omg, yeah why not :)) :D

packages/components/src/core/NavigationJumpTo/index.tsx
masoudmanson: Of course :D

tihuan (approved):
Amazing work! Thanks so much for finding the clever solution, @masoudmanson 🚀 !!

PR #465, opened 2023-05-17
refactor: tag screenshot feedback

tihuan:
## Summary

Github issue: [#XXXX](link)
Copy isuue descirption here

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -74,6 +74,7 @@ module.exports = {
           "**/*.stories.ts",
           "**/*.stories.tsx",
           "**/*.test.tsx",
+          "**/*.testUtils.tsx",
         ],
         // specify the path to the folder containing package.json
         packageDir: "./",
--- a/packages/components/src/core/Tag/index.stories.tsx
+++ b/packages/components/src/core/Tag/index.stories.tsx
@@ -12,41 +12,22 @@ const Tag = (props: Args): JSX.Element => {
   return <RawTag label={label} {...props} />;
 };
 
-const CUSTOM_COLOR_TUPLES = {
-  labelAndBack: ["#000000", "#C65FA7"],
-  labelAndBackAndIcon: ["#000000", "#C65FA7", "#FFD400"],
-};
-
-const availableColorOptions = [
-  "primary",
-  "info",
-  "success",
-  "warning",
-  "error",
-  "gray",
-  "beta",
-  CUSTOM_COLOR_TUPLES.labelAndBack,
-  CUSTOM_COLOR_TUPLES.labelAndBackAndIcon,
-];
-
-const availableIconOptions = [
+const ICONS = [
   undefined,
   <Icon sdsSize="l" sdsIcon="checkCircle" sdsType="button" />,
   <Icon sdsSize="l" sdsIcon="loading" sdsType="button" />,
   <WbSunny />,
   <CheckCircleOutline />,
 ];
-
 const SDS_STYLES: ExtraTagProps["sdsStyle"][] = ["rounded", "square"];
-const COLORS = availableColorOptions.slice(0, 9) as ExtraTagProps["tagColor"][];
-
-// FEEDBACK
-// FEEDBACK
-// FEEDBACK
-// This mutates the constant array, so it's no longer constant
-// Why do we need to dynamically remove the last color?
-COLORS.splice(7, 1);
-
+const COLORS: ExtraTagProps["tagColor"][] = [
+  "primary",
+  "info",
+  "error",
+  "gray",
+  "beta",
+  ["#000000", "#C65FA7", "#FFD400"],
+];
 const SDS_TYPES: ExtraTagProps["sdsType"][] = ["primary", "secondary"];
 
 export default {
@@ -61,13 +42,12 @@ export default {
           "error",
           "gray",
           "beta",
-          "Custom colors for Label, Background",
           "Custom colors for Label, Background, Icon",
         ],
         type: "select",
       },
-      mapping: availableColorOptions,
-      options: Object.keys(availableColorOptions),
+      mapping: COLORS,
+      options: Object.keys(COLORS),
     },
     hover: {
       control: { type: "boolean" },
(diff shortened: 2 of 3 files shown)


.eslintrc.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
           "**/*.stories.ts",
           "**/*.stories.tsx",
           "**/*.test.tsx",
+          "**/*.testUtils.tsx",
```
tihuan, line 77: allow `*.testUtils.tsx` to import devDependencies

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -38,13 +39,15 @@ const availableIconOptions = [
 
 const SDS_STYLES: ExtraTagProps["sdsStyle"][] = ["rounded", "square"];
 const COLORS = availableColorOptions.slice(0, 9) as ExtraTagProps["tagColor"][];
+
+// FEEDBACK
+// FEEDBACK
+// FEEDBACK
+// This mutates the constant array, so it's no longer constant
+// Why do we need to dynamically remove the last color?
```
tihuan, line 47: PTAL thank you!

packages/components/src/core/Tag/index.testUtils.tsx
```diff
@@ -0,0 +1,285 @@
...
+  <CheckCircleOutline />,
+];
+
+const iconOptions = AVAILABLE_ICON_OPTIONS.slice(0, 2);
+
+export const HOVER_OPTIONS = [true, false];
+
+export function CommonScreenshotTestDemo({
+  props,
+  colors,
```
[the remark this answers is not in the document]
liaprins: Well one other difference is that Gray is the only one whose `sdsType` is split between the main `ScreenshotTest` (gets gray x secondary) and its own `GrayPrimaryScreenshotTest` (gets gray x primary). I realized that some of the secondary Tags (with state of hover) for Success and Warning were also failing a11y test, so I put both primary and secondary of Success and Warning into what had been called `SuccessWarningPrimaryScreenshotTest` and renamed it to `SuccessWarningScreenshotTest` (without `Primary` in the name now).

packages/components/src/core/Tag/index.stories.tsx
liaprins: Well I didn't want to duplicate the color list since it was already declared as a const for use in this file in other tests (e.g. the Default story where user can change the Tag's options from the control panel). This list has two custom color palettes, one with two colors (for if you're not using an icon I guess, based on its name of `labelAndBack`) and one with three colors for it you are using an icon (`labelAndBackAndIcon`). 

However, in the `ScreenshotTest` permutations, each color goes through Tags both with and without icons. the `labelAndBackAndIcon` color palette will just render its first 2 colors if there's no icon, but the `labelAndBack` color palette will leave the icon blank/white since there's not a specific color included for it. Otherwise these two custom color palettes are identical and the permutations they render are redundant (other than the lack of color for icon), so I thought we'd only need to include `labelAndBackAndIcon` palette (since it automatically does the right colors even if there's not an icon). So anyway, this line is excluding the custom color palette that doesn't include a color for icons!

However, I'm not sure if the original color list that `COLORS` copies shoudl instead just be updated to only have the color palette that does include the icon color?

packages/components/src/core/Tag/index.testUtils.tsx
```diff
@@ -0,0 +1,285 @@
...
+  };
+
+  return (
+    <div style={DISPLAY_CONTENTS}>
+      <h3 style={h3Label}>
+        Color: <b>{color}</b>
+      </h3>
+      {SDS_TYPES.map((sdsType) => {
+        return (
+          <PrimaryTagType
```
tihuan, line 130: should we rename `PrimaryTagType ` to just `TagType`, since the loop goes through all `SDS_TYPES`, not just `primary`?

packages/components/src/core/Tag/index.testUtils.tsx
```diff
@@ -0,0 +1,285 @@
...
+}: {
+  props: Args;
+  colors: ExtraTagProps["tagColor"][];
+}): JSX.Element {
+  // loop through all SDS_STYLES
+  return (
+    <>
+      {SDS_STYLES.map((sdsStyle) => {
+        return (
+          <PrimaryTagStyle
```
tihuan, line 51: should we rename `PrimaryTagStyle` to just `TagStyles`, since we are going through all `SDS_STYLES`?

packages/components/src/core/Tag/index.stories.tsx
```diff
@@ -517,186 +248,18 @@ export const ScreenshotTest = {
...
+
 function GrayPrimaryScreenshotTestDemo(props: Args): JSX.Element {
-  // loop through all SDS_STYLES
   return (
-    <>
-      {SDS_STYLES.map((sdsStyle) => {
-        return <GrayPrimaryTagStyle sdsStyle={sdsStyle} key={sdsStyle} />;
-      })}
-    </>
+    <CommonScreenshotTestDemo props={props} colors={GRAY_PRIMARY_COLORS} />
```
tihuan, line 255: this is wrong, because it iterates both gray + primary and gray + secondary?

packages/components/src/core/Tag/index.testUtils.tsx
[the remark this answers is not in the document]
tihuan: Ah I see! So my refactored `GrayPrimaryScreenshotTestDemo` is wrong, because it iterates with gray + primary and gray + secondary, and we only want gray + primary?

If so, it sounds like we need to add `styles` prop to this component too?

packages/components/src/core/Tag/index.testUtils.tsx
liaprins: Yeah I think so!

packages/components/src/core/Tag/index.testUtils.tsx
liaprins: Yeah good point! I can make these updates after I merge this in!

liaprins (approved):

Issue #461, opened 2023-05-11
Chromatic test for Tooltip


liaprins:
### Alignment
- [ ] Align with design (and product, other engineers, if necessary) on the visual traits to include as permutation dimensions for the Chromatic test
- Find the component in this doc as a starting point: [Chromatic tests: Which component dimensions to consider looping through](https://docs.google.com/document/d/1i40YV1rX61dNzGsqbhqJAD8PL1puCecX7rnHgD_FawE/edit#)
- Permutation dimensions may include the component’s props, but it may not always be necessary to include every prop
- If the component can be interacted with, or if part of it can (e.g. an “x” button or similar within it), it should include pseudo state as a permutation dimension

- [ ] Once the permutation dimensions are aligned on, list them in this issue as checkboxes to track them

### Permutation dimensions to include
- [x] **Dimension**: properties

#### Exceptions, notes
- [ ] Add exceptions to combinations here
- ... and any other notes

### Tips for writing Chromatic tests
- Write the tests in index.stories.tsx within the component’s folder
- Name the story page LivePreview `export const LivePreview ... `
- Use or create an array per permutation dimension and loop through each to iteratively create all relevant combinations of the component
- Put pseudo-state at the lowest level
- See Button's Live Preview story for an example
- [More about Chromatic](https://storybook.js.org/docs/react/writing-tests/visual-testing#setup-chromatic-addon)
- [More about Storybook's Pseudo State addon](https://storybook.js.org/addons/storybook-addon-pseudo-states)

PR #472, opened 2023-05-20
test(Tooltip): Tooltip screenshot test

liaprins:
Updated Tooltip Storybook tests to have a Chromatic-specific test that iterates through all relevant visual variables.

#461

## Summary

Github issue: [#461 ](link)

- [x] **sdsStyle**: dark, light
- [x] **subtitle**: "sample text", undefined
- [x] **arrow**: true, false
- [x] **arrowOffset**: [any number between 120 and -120; I will use the following increments to test with : 120, 60, 0, -60, -120]
- [x] **width**: default, wide

#### Exceptions, notes
- [x] `sdsType=light` Tooltips cannot have `subtitle=true`
- [x] `sdsType=dark` Tooltips cannot have `width=wide`
- [x] `arrowOffset` can only apply to `arrow=true`
- Since the tooltip itself is not hoverable, but only appears on hover of another element, the Chromatic test should force that element to always be in hover state in order to show Tooltip

Here's what the new Screenshot Test story looks like:
Screenshot 2023-05-19 at 5 21 58 PM

## Checklist
- [x] Test Story in Storybook
- [x] Tests written
- [ ] Variables from `defaultTheme.ts` used wherever possible // liaprins: uncertain...
- [ ] If updating an existing component, depreciate flag has been used where necessary // liaprins: N/A
- [ ] Chromatic build verified by @chanzuckerberg/sds-design // liaprins: TBD, since this updates the Chromatic build by definition...


liaprins:
I just noticed that in the Chromatic screenshot the Tooltips are in some cases rendering so they have a placement above the parent component (info icon) which means they cover the labels describing which combination of visual variables they represent (especially for the dark-type Tooltips). I think I could try to fix this by assigning them all `placement="bottom"` like they are rendering by default on my screen, if this matters!

tihuan (approved):
Great test cases! Thanks for the screenshot too, even though I could technically go to Chromatics to see the result, it's still a lot faster to just see the screenshot in the PR ✨ 

PR is so good that I don't have anything to add 🔥 Great work!

PR #483, opened 2023-05-25
MenuItem Screenshot Test

liaprins:
## Summary

**MenuItem**
Github issue: #479 

### Still to fix:
- [ ] **`details` not rendering:**
- When I try to pass the MenuItem some details text as a prop called `details`, it’s not rendering any of that text (looks just like a MenuItem without details).
- I can’t tell if there’s a different format I need to use (other than passing it as a prop called `details`) since details is also not included in the Default or Test stories, and I couldn’t find anything about details in its index.tsx file…. 
- In the meantime, I can either still include details as a visual trait in the Chromatic screenshot test (it just won’t render properly until we know how to fix it), or I can remove it as a loop altogether (for now at least).

_Screenshot of the portion of the layout that should be rendering `details`, but is not:_
details-not-rendering

- [ ] **The `sdsIcon` prop is not recognizing the type I assigned it**
- I think because icons used within MenuItem are already assigned their own type (`IconName`) in the index.tsx file. But I cannot figure out how to transport that "type" into the Screenshot Test story... 
- I tried a similar approach to what Timmy helped me with, in the ButtonIcon Screenshot Test (not yet merged in, so will have to look at the PR for it), but I don't know TypeScript yet, so I just don't know what is going wrong with my attempts...

- [x] **isMultiSelect:** false, true
- [x] **details:** [sample details text], undefined
- [x] **columnValue:** [sample text], undefined
- [x] **sdsIconProps:** [include a sample icon], undefined
- [x] **selected:** false, true
- [x] **disabled:** false, true
- [x] **pseudo-state:** default, hover, active, focus

_Here's a screenshot of the current UI (most of it; it si too long to get it all in one screen!):_
screenshot

#### Exceptions
- [x] I will only render `disabled=true` if `pseudo-state=default`, since that is the only state wherein MenuItem could be disabled anyway, and this follows the pattern for how other Chromatic tests deal with disabled state, and how the states are shown in Zeroheight

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/src/core/MenuItem/index.stories.tsx
+++ b/packages/components/src/core/MenuItem/index.stories.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable no-use-before-define */
 import { Args, Meta } from "@storybook/react";
 import * as React from "react";
+import Icon from "../Icon";
 import RawMenuItem from "./index";
 import { DemoWrapper } from "./style";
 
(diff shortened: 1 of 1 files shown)


packages/components/src/core/MenuItem/index.stories.tsx
```diff
@@ -61,6 +61,316 @@ export const Default = {
...
+                    {disabled ? "Disabled: " : "State: "}
+                    <b>{disabled ? "true" : state}</b>
+                  </p>
+                  <RawMenuItem
+                    {...props}
+                    data-testid="menu-item"
+                    isMultiSelect={isMultiSelect}
+                    details={details}
+                    column={column}
+                    sdsIcon={sdsIcon}
```
liaprins, line 333: The `sdsIcon` prop is not recognizing the type I assigned it, I think because icons used within MenuItem are already assigned their own type (`IconName`) in the index.tsx file. But I cannot figure out how to transport that "type" into the Screenshot Test story... I tried similar approaches to what Timmy helped me with, in the ButtonIcon Screenshot Test (not yet merged in, so will have to [look at the PR for it](https://github.com/chanzuckerberg/sci-components/pull/462/files#r1192814051)), but I don't know TypeScript yet, so I just don't know what is going wrong with my attempts...

packages/components/src/core/MenuItem/index.stories.tsx
```diff
@@ -61,6 +61,316 @@ export const Default = {
...
+    return (
+      <div style={LEVEL_STYLE}>
+        <p style={LABEL_STYLE}>
+          Multi-select: <b>{isMultiSelect ? "true" : "false"}</b>
+        </p>
+        {DETAILS_OPTIONS.map((details) => {
+          return (
+            <MenuItemDetails
+              isMultiSelect={isMultiSelect}
+              details={details}
```
liaprins, line 127: `details` is not rendering. When I try to pass the MenuItem some details text as a prop called `details`, it’s not rendering any of that text (looks just like a MenuItem without details). I can’t tell if there’s a different format I need to use (other than passing it as a prop called `details`) since details is also not included in the Default or Test stories, and I couldn’t find anything about details in its index.tsx file….
In the meantime, I can either still include details as a visual trait in the Chromatic screenshot test (it just won’t render properly until we know how to fix it), or I can remove it as a loop altogether (for now at least).

tihuan:
@liaprins just pushed a commit to fix type error 🤩 That one was hard lol

packages/components/src/core/MenuItem/index.stories.tsx
tihuan: this should be resolved by my latest commit! Thank you!

PR #490, opened 2023-06-01
test(DropdownMenu): DropdownMenu screenshot test

liaprins:
## Summary

**DropdownMenu**
Github issue: #485 

- [x] **title:** [sample title text], undefined
- [x] **search:** false, true
- [x] **groupBy:** [no], [yes]
- [x] **component:** [`Tag` component], undefined

#### Exceptions, notes
- [x] Find out how `groupby` options are specified, and how to _not_ specify it 
- `details` is currently controlled by `DropdownMenu`, however it is not included in this test because it is going to soon be switched to be under the control / a prop of `MenuItem` instead.

_Here's a screenshot of the layout:_
Screenshot 2023-06-01 at 1 27 01 PM

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


liaprins:
Copying a comment from @tihuan that was left in the TagFilter Screenshot Test PR, but I think was actually supposed to be a comment in this PR:
> Looks like the DropdownMenu components are overlapping in the screenshot. Maybe we just need more spaces among the dropdowns?
> 
> Thank you!
> 
> Screenshot 2023-06-06 at 9 22 32 AM

I will make this change now! Hopefully it won't be messed up since I already clicked on "Update branch" (or something like that) button above, but I haven't merged it yet, so should be okay to still make changes

liaprins:
Oops I accidentally merged this one before making the update :( I will try to fix this by doing "Restore branch" and making the change...

liaprins:
EDIT: @tihuan I know why they are overlapping... bascially, the dropdowns automatically change whether they are above or below the input in order to ensure they are always visible on the screen. I didn't think it'd do it in the Chromatic screenshot, since I thought it took a pic of everything in the UI without any regard to a screensize... I will see if there's a way to force DropdownMenu to always be below.

liaprins:
> Oops I accidentally merged this one before making the update :( I will try to fix this by doing "Restore branch" and making the change...

Actually I will just make a new branch + PR to fix this. And I will be just adding a LOT of top-margin to each iteration,  to ensure they never overlap each other, regardless of whether they render on top or bottom!

Issue #501, opened 2023-06-07
Omit "nonce"-"rev"-"rel"-"autoFocus"-"content" from all SDS components


masoudmanson:
Support Channel Issue: https://czi-sci.slack.com/archives/C032S43KKFV/p1684536524081589

In recent observations, it has come to our attention that the addition of five new types ("nonce" | "rev" | "rel" | "autoFocus" | "content") to the SDS components has resulted in operational difficulties for the product teams. 

In order to mitigate the problems faced by the product teams, a comprehensive investigation was conducted to identify the underlying cause. After careful analysis, it was determined that temporarily omitting these types would allow the CZID team to proceed with using the updated version of SDS without encountering any hindrances.

It is important to note, however, that this temporary solution does not provide a permanent resolution to the problem. Therefore, it is crucial that we prioritize the resolution of this bug and restore the omitted types in order to rectify the issue effectively.

Up Next: #502

PR #503, opened 2023-06-07
fix(types): fix typescript types

masoudmanson:
## Summary

Github issue: #501 

Support Channel Issue: https://czi-sci.slack.com/archives/C032S43KKFV/p1684536524081589

In recent observations, it has come to our attention that the addition of five new types ("nonce" | "rev" | "rel" | "autoFocus" | "content") to the SDS components has resulted in operational difficulties for the product teams.

In order to mitigate the problems faced by the product teams, a comprehensive investigation was conducted to identify the underlying cause. After careful analysis, it was determined that temporarily omitting these types would allow the CZID team to proceed with using the updated version of SDS without encountering any hindrances.

It is important to note, however, that this temporary solution does not provide a permanent resolution to the problem. Therefore, it is crucial that we prioritize the resolution of this bug and restore the omitted types in order to rectify the issue effectively.

Up Next: https://github.com/chanzuckerberg/sci-components/issues/502

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


tihuan (approved):
Thanks so much for manually adding the Omit, Masoud 🙏 Great solution!

tihuan:
Ahaha my pleasure! You did all the hard work 🙏 Thank you for that!! Jumping in the next meeting now 😆

Issue #562, opened 2023-08-16
Fix InputDropdown's Focus/Focus:visible states


masoudmanson:
If the user clicks on the input and opens it, then the state changes from `default` to `open` and the `sdsStage` changes from `default` to `userInput`. If they make a selection from the dropdown menu, the menu closes, then the state on the inputDropdown changes back to `default` (since the menu is no longer open), but the `sdsStage` remains on userInput since they have made a selection from the menu and the input is no longer in a default state.

Here's the default state (state = default; sdsStage = default) 

![Image](https://github.com/chanzuckerberg/sci-components/assets/927990/4bac2cff-c2a6-4071-aa55-e4d9fa6c3fba)

Here's the open state (state = open; sdsStage = userInput)

![Image](https://github.com/chanzuckerberg/sci-components/assets/927990/8c4e2562-92d3-4bca-9a52-4aa33e762fee)

Here's the closed state after the user has made a selection from the menu (state = default; sdsStage = userInput)

![Image](https://github.com/chanzuckerberg/sci-components/assets/927990/cc1e888b-b4e6-44b6-8b20-184af5611c6d)

You can see how the text color is different when state = default vs sdsStage = userInput

PR #565, opened 2023-08-22
fix(inputdropdown): fix UserInput and Open state styles

masoudmanson:
## Summary

**InputDropdown**
Github issue: #562 

## Checklist

- [x] Test Story in Storybook
- [x] Tests written

### DIFF
--- a/packages/components/src/core/ComplexFilter/__snapshots__/index.test.tsx.snap
+++ b/packages/components/src/core/ComplexFilter/__snapshots__/index.test.tsx.snap
@@ -5,7 +5,7 @@ exports[`<ComplexFilter /> Default story renders snapshot 1`] = `
   class="css-1x0reya"
 >
   <button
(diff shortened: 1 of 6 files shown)


clarsen-czi:
Within Storybook, if users select multiple=true and sdsType=value, the component should be hidden and the following message should appear in its place: "The "Value" variant of InputDropdown cannot be set to "multiple"."

clarsen-czi (changes requested):
A few small changes needed as we just discussed on the call.

clarsen-czi (dismissed):
Almost there, but I'm getting an error when sdsType=label and multiple=true. I got this after trying a few other combinations including sdsType=value and multiple=true (which displays the warning we chatted about which looks great!).

clarsen-czi (approved):
LGTM, thanks @masoudmanson !!

PR #581, opened 2023-08-30
fix: Use named import for MUI components

tihuan:
## Summary

Fix breaking imports

Single Cell Team noticed that importing SDS Autocomplete throws the following error:

`Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.`

Screenshot 2023-08-30 at 9 00 00 AM

And the root cause seems to be SDS TS -> JS transpiler didn't take ESM syntax into account when transpiled the code to CJS, so it resulted in importing the whole module instead of the intended module default export. 

For example,

```ts
// packages/components/dist/index.cjs.js
var Autocomplete = require('@mui/material/Autocomplete');
```

`Autocomplete` is the whole module object, when we just want `Autocomplete.default`

```
// we got the whole export module
Autocomplete {
  autocompleteClasses: [Getter],
  createFilterOptions: [Getter],
  default: [Getter],  // <--------------- we just want this, which is what ESM syntax does, but not for CJS
  getAutocompleteUtilityClass: [Getter]
}
```

The solution is to avoid importing MUI components through their component directories and use `import { Autocomplete } from '@mui/material'` instead, like so:

```ts
import { Autocomplete } from "@mui/material";
```

This way the transpiled cjs file will have the correct export:

```ts
var material = require('@mui/material');
var Autocomplete = material.Autocomplete; // This is the default export component we want
```

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/rollup.config.mjs
+++ b/packages/components/rollup.config.mjs
@@ -11,11 +11,14 @@ const config = [
   {
     external: [
       ...Object.keys(pkg.peerDependencies || {}),
-      "@mui/material/Popper",
-      "@mui/material/ToggleButton",
       "@mui/material/styles",
-      "@mui/material/Autocomplete",
-      "@mui/material/ToggleButtonGroup",
+      /**
+       * (thuang): Do NOT import MUI components from their component directory directly, since
+       * it breaks the ts to js transpilation. Instead, import from "@mui/material"
+       * For example:
+       * BAD: import Autocomplete from "@mui/material/Autocomplete";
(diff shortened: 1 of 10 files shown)


masoudmanson (approved):
Thank you for fixing this, @tihuan. Our team is so fortunate to have your expertise :D 💯

tihuan:
Aww really appreciate your kind words, Masoud! Right back at you ❤️❤️❤️!!

Issue #601, opened 2023-09-13
Add Determinate/Indeterminate variant to the MenuItem component


masoudmanson:
As part of implementing the new Multi-column dropdown, we need to temporarily introduce a determinate/indeterminate variant for the MenuItem component. This change is temporary as we're planning to use the InputCheckbox component inside the MenuItem component to handle checkbox states instead of directly managing them within the MenuItem component.

PR #602, opened 2023-09-13
feat(menuitem): add determinate/indeterminate variant to the MenuItem

masoudmanson:
## Summary

**MenuItem**
Github issue: #601 

As part of implementing the new Multi-column dropdown, we need to temporarily introduce a determinate/indeterminate variant for the MenuItem component. This change is temporary as we're planning to use the InputCheckbox component inside the MenuItem component to handle checkbox states instead of directly managing them within the MenuItem component.

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


clarsen-czi (approved):
LGTM! Thanks @masoudmanson !!

package.json
```diff
@@ -110,9 +110,9 @@
     "build-storybook": "storybook build -o docs-build",
     "test-storybook": "test-storybook",
     "storybook:axe": "yarn build-storybook && yarn storybook:axeOnly",
-    "storybook:axeOnly": "axe-storybook --build-dir docs-build",
+    "storybook:axeOnly": "axe-storybook --build-dir docs-build --timeout 10000",
```
tihuan, line 113: Thanks for adding this, @masoudmanson !

We can probably remove the following config in test files now we set the flag globally!

```ts
axe: {
  timeout: 10 * 1000,
},
```

masoudmanson: Ah, you're right. 
We don't even need a global timeout config. I'll revert this change and set a timeout for that specific component. 😯😥

#611

PR #620, opened 2023-10-04
fix: call onChange in event callback instead of useEffect

tihuan:
## Summary

1. Remove `onChange()` call in useEffect and use event handlers instead. This is to align with React best practices ([here](https://react.dev/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects))
2. This resolves an infinite loop bug in Single Cell, since SC passes a new `onChange` function down to Dropdown every time a value changes
3. I have built the library and installed it locally to test it in SC and can confirm that the fix works 🎉 !

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/src/core/Dropdown/index.tsx
+++ b/packages/components/src/core/Dropdown/index.tsx
@@ -221,9 +221,7 @@ const Dropdown = <Multiple extends boolean | undefined = false>({
       }
     } else {
       if (multiple) {
-        setValueAndCallOnChange(
-          pendingValue as Value<DefaultDropdownMenuOption, Multiple>
-        );
+        setPendingValue(value as MUIValue<true>);
       }
 
       setAnchorEl(event.currentTarget);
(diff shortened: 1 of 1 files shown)


packages/components/src/core/Dropdown/index.tsx
```diff
@@ -221,7 +221,9 @@ const Dropdown = <Multiple extends boolean | undefined = false>({
       }
     } else {
       if (multiple) {
-        setPendingValue(value as MUIValue<true>);
```
masoudmanson, line 224: I think we shouldn't change this one cause it's a `SetPendingValue` and not a `SetValue`!

masoudmanson (approved):
Wow, so brilliant. Thank you for fixing this @tihuan 🥳😍
There's only one small change I believe you should revert. Otherwise, it LGTM.

packages/components/src/core/Dropdown/index.tsx
tihuan: Ooh yes thanks so much for catching this critical bug 😆 Really appreciate it! Reverted 🙆‍♂️

Issue #686, opened 2023-11-06
Typo in font weights 


cathystoli:
From Single cell: Seve found a typo in our defaultTheme file. fontWeights.medium should be 500 not 800.

https://github.com/chanzuckerberg/sci-components/blob/54b380d31dc834b27fa9567fe604cc924b5a3ce4/packages/components/src/core/styles/common/defaultTheme.ts#L13-L20

cathystoli:
@masoudmanson I am going to add this to on deck for our next sprint

PR #707, opened 2023-11-22
fix(defaulttheme): fix typo in fontWeights, medium = 500

masoudmanson:
## Summary

**DefaultTheme**
Github issue: #686

Typo fixed in `defaultTheme` FontWeights object. Medium font-weight is equal to `500` now.


tihuan (approved):
Thanks for the fix!

PR #708, opened 2023-11-22
docs(dropdownmenu): add @deprecated tag to comment

masoudmanson:
## Summary

**DropdownMenu**

prefix the deprecation comment with a @deprecated tag so we get the strikethrough text to encourage users to migrate to the new type.

```javascript
/**
 * @deprecated
 * (masoudmanson): We've replaced DefaultDropdownMenuOption with DefaultAutocompleteOption
 * as the preferred choice. However, for backward compatibility, we've exported this line to
 * prevent potential TypeScript type issues for product teams using previous versions.
 */
export type DefaultDropdownMenuOption = DefaultAutocompleteOption;
```

### DIFF
--- a/packages/components/src/core/DropdownMenu/index.tsx
+++ b/packages/components/src/core/DropdownMenu/index.tsx
@@ -22,11 +22,11 @@ import {
 } from "./style";
 
 /**
+ * @deprecated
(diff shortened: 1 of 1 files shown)


tihuan (approved):
Hurray thank you! I forgot about this too 😆 Thanks for the quick followup PR!

PR #712, opened 2023-11-29
SDS Hackday: test setup + accordion tests

ahuth:
This PR does a couple things:
- Removes the unused `jest-canvas-mock` and `eslint-plugin-jest-playwright` deps
- Adds `@testing-library/jest-dom` so we can use its matchers
- Makes the Accordion tests more comprehensive

I'm planning on going through more of the tests, but wanted to stop here and get the test setup and DOm matchers in.


.eslintrc.js
```diff
@@ -15,7 +15,6 @@ module.exports = {
     "plugin:jsx-a11y/recommended",
     "airbnb-typescript",
     "plugin:sonarjs/recommended",
-    "plugin:jest-playwright/recommended",
```
ahuth, line 18: As far as I can tell, there aren't any playwright tests.

jest.common.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   moduleDirectories: ["node_modules", "<rootDir>"],
   preset: "ts-jest",
-  setupFiles: ["jest-canvas-mock", "../../intersectionObserverMock.js"],
+  setupFilesAfterEnv: ["./jest.setup.ts"],
```
ahuth, line 4: Removed the unused jest-canvas-mock dep.

Also moved the intersectionObserver mock and the testing-library/jest-dom setup into a setup file (which makes the typescript integration easier).

packages/components/src/core/Accordion/index.test.tsx
```diff
@@ -1,29 +1,40 @@
...
-import { fireEvent, render, screen } from "@testing-library/react";
-import * as snapshotTestStoryFile from "./index.stories";
-import Meta, { Test as TestStory } from "./index.stories";
+import { composeStories } from "@storybook/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import * as stories from "./index.stories";
 
 // Returns a component that already contain all decorators from story level, meta level and global level.
-const Test = composeStory(TestStory, Meta);
+const { Test } = composeStories(stories);
```
ahuth, line 7: Since we already have an import of the whole stories file, we can remove an import by using `composeStories` instead of `composeStory`.

packages/components/src/core/Accordion/index.test.tsx
```diff
@@ -1,29 +1,40 @@
...
 
 describe("<Accordion />", () => {
-  generateSnapshots(snapshotTestStoryFile);
+  generateSnapshots(stories);
 
   it("renders accordion component", () => {
-    render(<Test {...Test.args} />);
+    render(<Test />);
```
ahuth, line 13: No need to pass in the story's args. Using `composeStory` / `composeStories` already does that.

packages/data-viz/jest.setup.ts
```diff
@@ -0,0 +1,2 @@
+import "@testing-library/jest-dom";
+import "../../intersectionObserverMock";
```
ahuth, line 2: Looks like data-viz doesn't have any tests, so we could remove jest and its setup from it if we wanted to simplify things a bit.

.eslintrc.js
ehoops-cz: Are we adding any today?

ahuth: It's not mentioned in the spreadsheet.

Also, Playwright itself isn't in this repo. This is just an eslint plugin

packages/components/src/core/Accordion/index.test.tsx
```diff
@@ -1,29 +1,40 @@
...
+    fireEvent.click(accordionHeader);
+    expect(accordionHeader).toHaveAttribute("aria-expanded", "true");
+    expect(screen.getByText(/Lorem ipsum/)).toBeVisible();
+
+    // And closes when clicked again.
+    fireEvent.click(accordionHeader);
+    expect(accordionHeader).toHaveAttribute("aria-expanded", "false");
+
+    await waitFor(() => {
+      expect(screen.getByText(/Lorem ipsum/)).not.toBeVisible();
```
ehoops-cz, line 37: Why does this line need the waitFor when the lines above do not need it?

ehoops-cz (approved):
Looks good!  Left a couple questions.

.eslintrc.js
tihuan: That sounds good! With Chromatics, maybe there's no immediate need to use Playwright. But if we do, we can add back later 👍

jest.common.config.js
tihuan: Oh great idea thanks for doing this!

packages/components/src/core/Accordion/index.test.tsx
```diff
@@ -1,29 +1,40 @@
...
-    fireEvent.click(accordionElement);
-    expect(
-      screen.getAllByText(
-        "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse malesuada lacus ex, sit amet blandit leo lobortis eget."
-      )
-    ).not.toBeNull();
+
+    // Starts closed
+    const accordionHeader = screen.getByRole("button");
+    // eslint-disable-next-line sonarjs/no-duplicate-string
```
tihuan, line 23: nit: should we extract duplicate string as a constant 🙏 ?

packages/components/src/core/Accordion/index.test.tsx
```diff
@@ -1,29 +1,40 @@
...
+    expect(accordionHeader).toHaveAttribute("aria-expanded", "true");
+    expect(screen.getByText(/Lorem ipsum/)).toBeVisible();
+
+    // And closes when clicked again.
+    fireEvent.click(accordionHeader);
+    expect(accordionHeader).toHaveAttribute("aria-expanded", "false");
+
+    await waitFor(() => {
+      expect(screen.getByText(/Lorem ipsum/)).not.toBeVisible();
+    });
```
tihuan, line 38: so much better 🔥 Thank you!

packages/data-viz/jest.setup.ts
tihuan: I think @masoudmanson will start adding tests in December for data-viz Heatmap, so let's keep it 💪 Thank you!

tihuan (approved):
LGTM! Thanks for creating the first SDS Hackday PR, @ahuth 🔥 🏆 

Only one tiny non-blocking nit!

packages/components/src/core/Accordion/index.test.tsx
ahuth: It seems like going from expanded back to collapsed is async for some reason, and takes a couple ms to happen.

Not sure why it's different than the collapsed -> expanded case.

packages/components/src/core/Accordion/index.test.tsx
ahuth: Sure, will do

ahuth:
Thanks 🙏

PR #714, opened 2023-11-29
fix: add focused stylign to tabs

ahuth:
This PR adds a focus outline to the tabs component, for the [SDS accessibility hackday project](https://docs.google.com/spreadsheets/d/1NrbRZ4vTfeYtKU3GXrsmeS6QSHsNSwiV_ITLvo251WA/edit#gid=915097901).

Screenshot 2023-11-29 at 12 56 57 PM

[See it live on Chromatic](https://www.chromatic.com/component?appId=616981d9028812003ade73f3&csfId=tabs&buildNumber=1592&k=65677c0a9bbad7e409e35add-1200px-interactive-true&h=8&b=-1)

Keyboard usage of the tabs is actually working as expected:
- Tab onto and off of the active tab
- When focused on the active tab, left and right arrows changes the active tab
- Space changes to the selected tab

### DIFF
--- a/packages/components/src/core/Tabs/style.tsx
+++ b/packages/components/src/core/Tabs/style.tsx
@@ -94,6 +94,11 @@ export const StyledTab = styled(RawTab, {
         color: black;
       }
 
+      &.Mui-focusVisible {
+        outline: 5px auto Highlight;
+        outline: 5px auto -webkit-focus-ring-color;
+      }
+
       &.Mui-selected {
         color: black;
 
--- a/packages/components/src/core/Tabs/__snapshots__/index.test.tsx.snap
+++ b/packages/components/src/core/Tabs/__snapshots__/index.test.tsx.snap
@@ -14,7 +14,7 @@ exports[`<Tabs /> Default story renders snapshot 1`] = `
     >
       <button
         aria-selected="true"
-        class="MuiButtonBase-root MuiTab-root MuiTab-textColorPrimary Mui-selected css-17rzz1x-MuiButtonBase-root-MuiTab-root"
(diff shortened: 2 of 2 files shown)


packages/components/src/core/Tabs/style.tsx
```diff
@@ -94,6 +94,11 @@ export const StyledTab = styled(RawTab, {
         color: black;
       }
 
+      &.Mui-focusVisible {
+        outline: 5px auto Highlight;
+        outline: 5px auto -webkit-focus-ring-color;
```
ahuth, line 99: This is currently the same focus outline that buttons have:
Screenshot 2023-11-29 at 12 56 57 PM

I am not a designer, and I'm happy to change this to something else. Figured it made sense for now to keep it simple and match buttons and text inputs.

Note that mui has default focus styles which we're suppressing as part of the theme: https://github.com/chanzuckerberg/sci-components/blob/9451847d508ead93707839e488da799c629914ef/packages/components/src/core/styles/common/defaultTheme.ts#L283-L298

We could turn those back on, although it looks weird and would take tweaking colors for both buttons and tabs.

ehoops-cz: 🤔 hmm... Are we re-doing work that we can just get for free from Mui?

ehoops-cz (approved):

packages/components/src/core/Tabs/style.tsx
ahuth: We definitely are, although I think it would take design and eng work to get back to something that looks reasonable with the defaults for all components.

Issue #577, opened 2023-08-25
`icon={false}` not rendering as expected for Callout component


liaprins:
**Describe the bug**
Setting the `icon` prop for the Callout component to `{false}` should remove the icon, but it does not.

**To Reproduce**
1. Add `icon={false}` as a prop to the Callout component (e.g. in a Storybook test, where it can easily be checked)

**Expected behavior**
This should remove the icon, like it does for the [MUI component it is built upon](https://mui.com/material-ui/react-alert/#icons). I also verified with Timmy that this is the expected behavior; see more in a Zeroheight comment [here](https://czi.zeroheight.com/styleguide/s/66099/p/72e266/c/100679).

**Recommended fix (per @tihuan ), documenting here for posterity:**
Change the [source code](https://github.com/chanzuckerberg/sci-components/blob/bdb41f9a1e4581e45d9adeda7633593585f717e0/packages/components/src/core/Callout/index.tsx#L59) from

`if (icon) return icon;`

to:

`if (icon !== undefined) return icon;`

"And then the following will work:
icon={false} - it will not take up space in the component. Similar to display: none;
icon={<></>} - it will still take up space in the component, but just empty space. Similar to visibility: hidden;
icon={null} - this is a little counter intuitive, but it will fall back to using MUI's iconMapping" 
(Thanks, @tihuan !!!)

PR #717, opened 2023-11-29
fix(callout): allow Callout component to not have an icon

liaprins:
Callout component can now exist sans icon by adding icon={false} as a prop

re #577

## Summary

**Callout**
Github issue: #577 

Recommended fix (per @tihuan ), documenting here for posterity:
Change the [source code](https://github.com/chanzuckerberg/sci-components/blob/bdb41f9a1e4581e45d9adeda7633593585f717e0/packages/components/src/core/Callout/index.tsx#L59) from

`if (icon) return icon;`
to:
`if (icon !== undefined) return icon;`
"And then the following will work:
icon={false} - it will not take up space in the component. Similar to display: none;
icon={<></>} - it will still take up space in the component, but just empty space. Similar to visibility: hidden;
icon={null} - this is a little counter intuitive, but it will fall back to using MUI's iconMapping"

## Checklist

- [x] Test Story in Storybook
- [x] Tests written

**Note that I am also having David Ruiz  help with reviewing bug fixes for SDS Hackday (which fixing this bug was part of), but I couldn't find him listed yet when I went to add reviewers**


DRuizCZI:
LGTM but i will feel better after @timmy or @masoudmanson  Take a look

packages/components/src/core/Callout/index.tsx
```diff
@@ -56,7 +56,7 @@ const Callout = ({
   };
 
   const getIcon = () => {
-    if (icon) return icon;
+    if (icon !== undefined) return icon;
```
tihuan, line 59: Niice! Thanks so much for the fix, @liaprins 🙏 ❤️ 

Can we add a comment to explain why we explicitly use `icon !== undefined` here please? In case someone in the future "refactors" this back to `if (icon) return icon` lol

Something like:

```ts
/**
 * (liaprins): Explicitly use `icon !== undefined` to fix bug in ticket: https://github.com/chanzuckerberg/sci-components/issues/577 
 */
```

Or something better!

masoudmanson (approved):
Great catch, @liaprins! 
Thank you for addressing this issue. The change LGTM. 

However, before merging, let's ensure we add the comment mentioned by @tihuan. Once that's added, we're all set to merge this. Appreciate your quick action on this! 🎉⭐️

packages/components/src/core/Callout/index.tsx
liaprins: Great, thanks for the suggestion! I have finally updated to add the comment as suggested!

PR #718, opened 2023-11-29
feat: Dark Mode

lanesawyer:
## Dark Mode

Github issue: [#XXXX](link)
Copy isuue descirption here

## Remaining Work
- [ ] Real dark mode colors
- [ ] Checking all the components to make sure nothing is broken
- [ ] Accessibility verification
- [ ] Updated documentation for setting up the project since you'll need to provide a theme now (or, we just default to light or something)
- [ ] Probably other things I'm forgetting

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


.gitignore
```diff
@@ -108,3 +108,12 @@ storybook-static
 
 # OS Files
 .DS_Store
+
```
lanesawyer, line 111: Turns out I'm on the latest version of `yarn` which hasn't been used by this project yet so I was getting _tons_ of added files from `yarn`'s cache. This ensures the latest `yarn` doesn't cause `git` annoyances in the future 😄

packages/components/src/core/ButtonIcon/index.stories.tsx
```diff
@@ -95,7 +95,7 @@ export const Default = {
 // Live Preview
 
 const LivePreviewDemo = (): JSX.Element => {
-  const spacings = defaultAppTheme?.spacing;
+  const spacings = sharedAppTheme?.spacing;
```
lanesawyer, line 98: Probably okay? But there might be a better way to access the theme in the story that we could use without having to import it directly.

packages/components/src/core/Callout/style.ts
```diff
@@ -39,12 +38,13 @@ export const StyledCallout = styled(Alert, {
     // any buttom margin
     const titleBottomMargin = props.collapsed ? "margin-bottom: 0;" : "";
 
+    // TODO: Theme shouldn't be saying it might be null
```
lanesawyer, line 41: Type issue, needs to be addressed.

packages/components/src/core/LoadingIndicator/index.stories.tsx
```diff
@@ -30,7 +30,7 @@ export const Default = {
 // Live Preview
 
 const LivePreviewDemo = (props: Args): JSX.Element => {
-  const spacings = defaultAppTheme?.spacing;
+  const spacings = sharedAppTheme?.spacing;
```
lanesawyer, line 33: Another spot where we want to see if we can consume the theme from something other than an import.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -75,8 +75,64 @@ const defaultThemeColors = {
   },
 };
 
-export const defaultAppTheme: AppTheme = {
-  colors: defaultThemeColors,
+const darkThemeColors = {
```
lanesawyer, line 78: Gross ChatGPT generated ones, will be replaced.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -75,8 +75,64 @@ const defaultThemeColors = {
...
+  warning: {
+    "100": "#302B23",
+    "200": "#382E24",
+    "400": "#F5A623",
+    "500": "#D8921F",
+    "600": "#946314",
+  },
+};
+
+export const sharedAppTheme: Omit<AppTheme, "colors" | "mode"> = {
```
lanesawyer, line 135: Not sure if I like this typing approach yet... At the very least explain the theme building system in comments here.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -97,6 +153,7 @@ export const defaultAppTheme: AppTheme = {
     xl: { height: 32, width: 32 },
     xs: { height: 10, width: 10 },
   },
+  // TODO: Dark mode?
```
lanesawyer, line 156: Do the shadows need to pull alphas from a dark mode?

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -239,40 +296,52 @@ export const defaultAppTheme: AppTheme = {
...
+
 // (mlila) whenever our theme uses colors, we need to make sure we allow consuming
 // applications to override those colors using their own custom theme.
 // By defining borders using defaultAppTheme.colors instead of defaultThemeColors,
 // we allow other apps to specify their colors once, and have them apply
-// throughtout the application, such as in borders, etc without having to manually
+// throughout the application, such as in borders, etc without having to manually
 // override every theme property that makes use of colors.
-defaultAppTheme.borders = {
+lightAppTheme.borders = {
```
lanesawyer, line 317: I didn't do this for `darkAppTheme` yet because I'm not sure the best approach here. Just a lot of copy/pasting. Might want some simple functions to generate the objects based on the theme passed in?

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -557,4 +633,5 @@ export interface Borders {
   warning: Border;
 }
 
-export const defaultTheme = createThemeAdaptor(defaultThemeOptions);
+export const defaultTheme = (theme: "light" | "dark") =>
```
lanesawyer, line 636: Comment explaining the setup. This means any `defaultTheme` usage is now a function, not an object. Has consequences for developer usage.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -418,7 +487,12 @@ export function makeThemeOptions(appTheme: AppTheme): SDSThemeOptions {
   };
 }
 
-const defaultThemeOptions: SDSThemeOptions = makeThemeOptions(defaultAppTheme);
+const chooseTheme = (theme: "light" | "dark") => {
```
lanesawyer, line 490: Is there more than just `light` and `dark`? Should this be an enum with the ability to expand to more than just these two? 🤔

packages/components/src/core/styles/common/defaultTheme.ts
lanesawyer, line 636: Actually, we should preserve the `defaultTheme` and use `light` mode so that the API doesn't cause breaking changes to consumers of the library. We can export a `theme` variable that is the function to determine which one to choose.

lanesawyer: Done!

packages/components/src/core/styles/common/defaultTheme.ts
lanesawyer: Done!

README.md
```diff
@@ -94,7 +94,7 @@ export const Title - styled(Typography)`
 ```ts
 import { css, SerializedStyles } from "@emotion/react";
 import { styled } from '@mui/material/styles';
-import { getColors, getCorners } from "@czi-sds/components";
+import { getColors, getSpaces } from "@czi-sds/components";
```
lanesawyer, line 97: Just fixing a typo in the README

packages/components/src/core/LoadingIndicator/index.stories.tsx
lanesawyer: Done!

packages/components/src/core/styles/common/defaultTheme.ts
lanesawyer: Explained!

packages/components/src/core/styles/common/defaultTheme.ts
lanesawyer: Nope, for now they're okay the way they are for both themes.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -437,6 +511,8 @@ interface AppTheme {
   shadows: Shadows;
   spacing: Spacings;
   typography: Typography;
+  // TODO: Do I like this approach?
+  mode: "light" | "dark";
```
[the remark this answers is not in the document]
lanesawyer: No, I don't think we should do that. Code changed to remove it.

packages/components/src/core/styles/common/defaultTheme.ts
lanesawyer: Using the MUI type! Done.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -557,4 +730,20 @@ export interface Borders {
   warning: Border;
 }
 
-export const defaultTheme = createThemeAdaptor(defaultThemeOptions);
+/**
+ * Default theme, uses light mode with no option to change it.
+ *
+ * @deprecated Use the `theme` to get a flexible light/dark mode theme function
```
lanesawyer, line 736: Maybe we don't deprecate it, some folks may never be interested in light/dark so why complicate their lives?

Alternatively, deprecate but give the new one a default theme of "light". 🤔

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -505,6 +637,9 @@ export interface Spaces {
 export type Spacings = Spaces;
 
 export interface Color {
+  900?: string;
+  800?: string;
+  700?: string;
```
lanesawyer, line 642: If we don't want all `Color` types to be able to climb this high we could make an `ExtendedColor` `interface` that extends `Color` and adds these three. Then it can be applied only where these values would be desired.

packages/components/src/core/DialogTitle/style.ts
```diff
@@ -24,7 +24,11 @@ export const StyledDialogTitle = styled(DialogTitle)`
 export const Title = styled(Typography)`
   ${fontHeaderXl}
 
-  color: black;
+  ${(props) => {
+    return `
+      color: ${props.theme.palette.text.primary};
```
lanesawyer, line 29: Just thinking about this a bit... `Typography` should be theme-aware. I think we might be able to get away with removing `color` entirely!

tihuan: Thanks for paving the way for our components to be runtime theme aware 💡 !! 

Yeah using semantic `text.primary` sounds like best practice to me too 👍 !

packages/components/src/core/Tabs/style.tsx
```diff
@@ -51,7 +51,9 @@ export const StyledTabs = styled(TempTabs, {
     return `
       margin-top: ${isLarge ? spaces?.l : spaces?.m}px;
       margin-bottom: ${isLarge ? spaces?.xl : spaces?.m}px;
-      border-bottom: ${underlined ? `2px solid ${colors?.gray[200]};` : "none"};
+      border-bottom: ${
+        underlined ? `2px solid ${props.theme.palette.divider};` : "none"
```
tihuan, line 55: super nice!

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -239,44 +252,193 @@ export const defaultAppTheme: AppTheme = {
   },
 };
 
+const lightAppTheme: AppTheme = {
+  colors: lightThemeColors,
+  ...sharedAppTheme,
```
tihuan, line 257: should we swap the order here, so we ensure `colors: lightThemeColors,` will overwrite `sharedAppTheme.colors` if `sharedAppTheme` happens to have `colors` at runtime 😄 ?

packages/components/src/core/styles/common/defaultTheme.ts
tihuan: Great question! Yeah I also feel that maybe don't deprecate it and just default to "light" sounds less intrusive 😄 👍

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -557,4 +692,20 @@ export interface Borders {
...
 }
 
-export const defaultTheme = createThemeAdaptor(defaultThemeOptions);
+/**
+ * Default theme, uses light mode with no option to change it.
+ *
+ * @deprecated Use the `theme` to get a flexible light/dark mode theme function
+ */
+export const defaultTheme = createThemeAdaptor(
+  makeThemeOptions(chooseTheme("light"), "light")
```
tihuan, line 701: feels like potentially `defaultTheme()` could call `theme()` under the hood?

Something like:

```ts
export const defaultTheme = theme("light");
```

What do you think 🙆‍♂️ ?

tihuan (commented):
Looks SUPER good! Thanks so much for going through the code to find places where we can start adopting runtime theme swapping!

These examples will definitely help us overhaul the rest of the stories and components in Q2 👏 

Really appreciate it!

Just some minor questions/feedbacks, otherwise LGTM 🚀 !!

packages/components/src/core/styles/common/defaultTheme.ts
lanesawyer: Good point! Even though I've said `sharedAppTheme` omits `colors`, that doesn't mean it'll actually happen.

The joys of TypeScript lol

Swapping!

packages/components/src/core/DialogTitle/style.ts
lanesawyer: Cool, going to remove the color entirely for `Title` then. Keeping for `SubTitle` since that won't be a secondary color by default!

packages/components/src/core/Callout/style.ts
lanesawyer: So this TODO is a little right and a little wrong, looking at it again.

The `props` here are listed as `CalloutProps`, which extends from the `CommonThemeProps` where `theme` is undefined.

At runtime, `theme` would never be undefined, so that's why I dropped the initial TODO here.

I'm sure we could do a bit of type juggling we could do to make it defined, but I think we'd want to do that in other components that are extending `CommonThemeProps` directly.

Given this PR is more targeted at making dark mode work, I'm inclined to live with the untrue `?` on `theme` for now.

packages/components/src/core/styles/common/defaultTheme.ts
```diff
@@ -239,44 +252,193 @@ export const defaultAppTheme: AppTheme = {
...
+    "400": `1px solid ${lightAppTheme.colors.warning[400]}`,
   },
 };
 
-export function makeThemeOptions(appTheme: AppTheme): SDSThemeOptions {
+const lightThemePalette = (appTheme: AppTheme): PaletteOptions => ({
+  divider: appTheme.colors.gray[200],
+  background: {
+    default: common.white,
+    secondary: appTheme.colors.gray[100],
```
lanesawyer, line 338: I think this was added by @hthomas-czi during the hackathon, it's throwing a type error because we don't have it on `TypeBackground`.

@hthomas-czi Is a secondary background still something that's needed? There's `paper` that we could be using that already exists, so we'll have three background colors if we do use `secondary`.

If it's needed, it's easy to add to the type. I just want to check 😁

lanesawyer:
@tihuan Feedback addressed! I have an outstanding question for @hthomas-czi, but overall this PR should be pretty close to ready.

Also one more question, should we add details about the theme to the README in this PR, or update the README once the colors and polish are completed so people won't start thinking it's ready?

tihuan (approved):
Awesome! Thanks for addressing the feedback quickly, Lane 🫡 LGTM!

CC: @masoudmanson in case I missed something 😎🙏

Thanks again!

masoudmanson (approved):
Wow, @lanesawyer, this is truly fantastic!
A massive thank you for tackling the heavy lifting and crafting such a robust PR for us. Really appreciate it.
The PR looks great to me. I'll be continuing to update components and collaborate with the design team to fine-tune the colors and appearance of the components.
This is so exciting! 🥳

`Finally, I feel right at home as a programmer. We got a dark theme` 😱😍

packages/components/src/core/styles/common/defaultTheme.ts
hthomas-czi: > I think this was added by @hthomas-czi during the hackathon, it's throwing a type error because we don't have it on `TypeBackground`.
> 
> @hthomas-czi Is a secondary background still something that's needed? There's `paper` that we could be using that already exists, so we'll have three background colors if we do use `secondary`.
> 
> If it's needed, it's easy to add to the type. I just want to check 😁

Hey @lanesawyer! Sorry for the delay responding. For some horrible reason, I don't receive notifications for GitHub mentions. 

I'll leave it up to you, but `Paper` doesn't fully address the reason I added a secondary background. They're equivalent in dark mode. But in light mode, `Paper` is always white, while `background[secondary]` is light gray in light mode. This maintains a consistent appearance between modes. 

Thanks!

masoudmanson:
@lanesawyer, Thanks a ton for your work on the dark theme! Since a lot has changed since your PR, and there were too many conflicts, I created a new PR to pick up where you left off. 

You can check it out here: [Dark Mode Implementation](https://github.com/chanzuckerberg/sci-components/pull/815)

Really appreciate your contributions! 🎉🥳

lanesawyer:
Woooo love to see it! Thanks for running with what I had and finishing it all up. That final PR is a big one!

Excited to start playing with the official dark mode once it's shipped 😄

PR #719, opened 2023-11-29
fix: add live region to loading indicator

ahuth:
Add live region to loading indicator, so screen readers will read out "Loading" Tested with VoiceOver on Safari.

Note that this won't tell people when loading is done. For that we'd need a system or components outside of the loading indicator.

But I think this is reasonable for now.

### DIFF
--- a/packages/components/src/core/LoadingIndicator/__snapshots__/index.test.tsx.snap
+++ b/packages/components/src/core/LoadingIndicator/__snapshots__/index.test.tsx.snap
@@ -1,5 +1,33 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`<LoadingIndicator /> CustomAriaLabel story renders snapshot 1`] = `
+<div
+  class="css-1o5w4es"
+>
+  <div
+    class="css-1ft6wgl"
+  >
+    <svg
+      aria-hidden="true"
+      class="MuiSvgIcon-root MuiSvgIcon-fontSizeMedium css-25m3b5-MuiSvgIcon-root"
+      data-file-name="IconLoadingLarge"
+      data-testid="IconLoadingLarge"
+      fillcontrast="white"
+      focusable="false"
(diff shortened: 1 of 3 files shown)


packages/components/src/core/LoadingIndicator/index.tsx
```diff
@@ -8,13 +8,20 @@ import {
 export type LoadingIndicatorProps = Omit<
   RawLoadingIndicatorProps,
   "nonce" | "rev" | "rel" | "autoFocus" | "content"
->;
+> & {
+  "aria-label"?: string;
```
ehoops-cz, line 12: Should we require this?  I guess that would be a breaking change.  I want people to fall into the pit of success though.

ehoops-cz (approved):

packages/components/src/core/LoadingIndicator/index.tsx
ahuth: Not a bad idea, although I think it's fine to have it be optional because there's a default of "Loading".

I put this in here in case someone wants to customize.

packages/components/src/core/LoadingIndicator/__snapshots__/index.test.tsx.snap
```diff
@@ -1,5 +1,33 @@
...
+      data-file-name="IconLoadingLarge"
+      data-testid="IconLoadingLarge"
+      fillcontrast="white"
+      focusable="false"
+      viewBox="0 0 22 22"
+    />
+  </div>
+  <span
+    aria-label="Loading cats list..."
+    aria-live="polite"
```
tihuan, line 22: wow didn't know `aria-live` was a thing 🙏 TIL thank you!

tihuan (approved):
Looks amazing! Thanks so much for improving a11y here and having an example for the rest of the components to follow 💡 !

ahuth:
Thanks 🙏

PR #723, opened 2023-11-29
chore: update issue templates

jfoo1984:
* Remove amend component issue template
* Remove new component issue template
* Add request component issue template
* Add support issue template


tihuan (approved):
Wow nice the form format is just so much better 🙏 ✨ Thanks so much for discovering this and converting the templates over to use it!

Feels like a possible next step is to deprecate Slack workflows and use these forms instead there?

Thanks so much again, @jfoo1984 🚀 !! You're amazing!!

PR #736, opened 2024-01-17
fix(callout): fix callout expand in production

codemonkey800:
## Summary

Fixes issue where the Callout component content was not expanding correctly in prdouction. This is because the property `name` is obfuscated in production for some environments. Setting and checking for `displayName` should fix the issue because it is not obfuscated in production.

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/src/core/Callout/components/CalloutTitle/index.tsx
+++ b/packages/components/src/core/Callout/components/CalloutTitle/index.tsx
@@ -1,4 +1,5 @@
 import { AlertTitleProps } from "@mui/material/AlertTitle";
+import { CALLOUT_TITLE_DISPLAY_NAME } from "../../constants";
 import { StyledCalloutTitle } from "./style";
 
 const CalloutTitle = ({ children }: AlertTitleProps): JSX.Element => {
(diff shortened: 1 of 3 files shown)


packages/components/src/core/Callout/components/CalloutTitle/index.tsx
```diff
@@ -5,4 +5,7 @@ const CalloutTitle = ({ children }: AlertTitleProps): JSX.Element => {
   return <StyledCalloutTitle>{children}</StyledCalloutTitle>;
 };
 
+// Display name required to ensure Callout expand works correctly.
+CalloutTitle.displayName = "CalloutTitle";
```
tihuan, line 9: Can we create a constant for the string, so there's single source of truth 😃?

E.g., `const CALLOUT_TITLE_DISPLAY_NAME = "CalloutTitle` 

`CalloutTitle.displayName = CALLOUT_TITLE_DISPLAY_NAME`


`children[0]?.type?.displayName === CALLOUT_TITLE_DISPLAY_NAME;`

tihuan (approved):
Wow! Such a tricky bug 🫨 Really amazing that you found this root cause 🔥 Appreciate the PR, Jeremy!!

tihuan:
Awesome thanks so much for the quick update, @codemonkey800 🎉 !

Feel free to merge 🚀 !!

PR #737, opened 2024-02-01
feat(heatmap): Base Heatmap Built

masoudmanson:
## Summary

**Heatmap**
Github issue: #634 

**Technical steps:** 
- [x] Add single cell heatmap to the demo to make sure it is working correctly 
- [x] PNG/SVG download 
- [x] Update API 

**Design work:**
- [ ] @liaprins to think about the limit and experience for the full size view [3]

**Requirements:** 
- [x] Camera scroll view
- [x] Entire heatmap view (if the heatmap is under a certain size, pending step above)
- [x] outline when highlighting the cell 

## Testing the PR with the Heatmap Demo Project

To test the Pull Request (PR) using the heatmap demo project on your local machine, follow these steps:

1. Switch to the latest heatmap branch in the SDS codebase:

    ```js
    git pull
    git checkout 634-base-heatmap-build
    ```

2. Install dependencies:

    ```js
    yarn
    ```

3. Build the SDS packages:

    ```js
    yarn build
    ```

4. Clone the heatmap demo into your projects folder:

    ```js
    git clone https://github.com/masoudmanson/sds-heatmap
    ```

5. Install dependencies in the demo project:

    ```js
    yarn
    ```

6. Remove the old `@czi-sds/data-viz` package:

    ```bash
    yarn remove @czi-sds/data-viz
    ```

7. Install `@czi-sds/data-viz` from local files:
   - For Yarn 1.x:

     ```bash
     yarn add file:../sci-components/packages/data-viz
     ```

   - For Yarn 2.x:

     ```bash
     yarn add @czi-sds/data-viz@file:../sci-components/packages/data-viz
     ```

8. Run the demo in development mode:

    ```bash
    yarn dev
    ```

By following these steps, you can thoroughly test the PR using the heatmap demo project on your local machine.

Screenshot 2024-02-01 at 10 14 36 AM

## Checklist

- [x] Default Story in Storybook
- [x] LivePreview Story in Storybook
- [x] Test Story in Storybook
- [x] Tests written
- [x] Variables from `defaultTheme.ts` used wherever possible
- [x] If updating an existing component, depreciate flag has been used where necessary
- [x] Chromatic build verified by @chanzuckerberg/sds-design

PR #749, opened 2024-02-14
refactor: Heatmap chart instance management

tihuan:
## Summary

Github issue: [#XXXX](link)
Copy issue description here

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/data-viz/src/core/HeatmapChart/index.tsx
+++ b/packages/data-viz/src/core/HeatmapChart/index.tsx
@@ -10,7 +10,7 @@ import {
   useState,
 } from "react";
 
-import { EMPTY_OBJECT } from "../../common/utils";
+import { EMPTY_OBJECT } from "src/common/utils";
 import { useUpdateChart } from "./hooks/useUpdateChart";
 import { CreateChartOptionsProps } from "./hooks/utils";
 import { ChartContainer } from "./style";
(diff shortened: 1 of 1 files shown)


packages/data-viz/src/core/HeatmapChart/index.tsx
```diff
@@ -52,22 +52,42 @@ const HeatmapChart = forwardRef(
...
+     * (thuang): We need both a state and a ref to store the chart instance, so
+     * some hooks can opt out of re-rendering when the chart instance changes.
+     */
+    const [chart, setChart] = useState<ECharts | null>(null);
+    const chartRef = useRef(chart);
+
+    /**
+     * (thuang): Use this ref to store the onEvents prop to prevent
+     * unnecessary re-renders when the onEvents prop changes.
+     * NOTE: This implies that `onEvents` prop changes alone from the parent
```
masoudmanson, line 68: Could this potentially cause issues later on?  
I mean the fact that changing the `onEvents` won't trigger re-renders as you've mentioned!

masoudmanson (approved):
This is fantastic! I really appreciate your help with this. It looks good to me; however, I have one question, which I've added as a comment. Thanks again, @tihuan!

packages/data-viz/src/core/HeatmapChart/index.tsx
tihuan: Yeah I was debating about this too 🤔 Really a balance between UX (asking users to `useCallback()` and `useMemo()` to memoize their `onEvents` object, vs. performance (re-render every time `onEvents` changes or not) 

I think in this case it feels like the event handlers are less likely to change than data being passed, so I feel it's probably an okay tradeoff here

Also I think the current implementation in [this PR](https://github.com/chanzuckerberg/sci-components/pull/737/files/f9c6febf8af2ecb68f2f204ae5b759f011d6d817#r1489896229) also ignores all the prop changes except for `[echartsRendererMode]` as well, so changing `onEvents` also won't re-render the chart!

Issue #766, opened 2024-05-02
Fix Autocomplete defaultOpen and open behaviors


masoudmanson:
Alec Tarashansky:

> How do I prevent the popper from opening by default when loading the page?

```
      <Autocomplete<Entity, false, false, false>
        data-testid={CELL_GUIDE_CARD_SEARCH_BAR_TEXT_INPUT}
        label={CELL_GUIDE_SEARCH_BAR_LABEL}
        search
        PopperComponent={(popperProps) => (
          <StyledPopper {...popperProps} fullWidth={skinnyMode} />
        )}
```

> I tried controlling the open state in the custom Popper component I'm using (StyledPopper), but there are a couple problems with it.
> 1. I'm trying to manually control open state with InputBaseProps={{onFocus=...,onBlur=...}} but it's a bit hacky and there are some issues with clicking on the popper triggering the onBlur, etc. All things I could figure out but I'm wondering if there's a better way
> 2. When controlling open state, the popper opens with none of the options displayed (z index issue perhaps?) This doesn't happen when the open state is not being controlled. I'm also wondering if the columns count as poppers themselves and their open state must be state-controlled as well (does this mean I need a custom Paper component?)

---

> Ideally, it'd be great if we had an "openByDefault" property or something since the sds autocomplete component already seems to handle focus and blur internally quite well. alternatively, making the open property on the Autocomplete component functional again (which controls state of all poppers) would be convenient

PR #767, opened 2024-05-02
fix(autocomplete): fix controlled open state

masoudmanson:
## Summary

**Autocomplete**
Github issue: #766 

This PR addresses issue #766 by fixing the controlled open state behavior of the Autocomplete component for both single and multi-column Autocompletes.

### Changes Made
- Implemented a fix for the controlled open state behavior.
- Added handlers for both opening and closing the Autocomplete component.
- Demonstrated the usage of controlled open state with a simple example in the storybook (`Autocomplete/Controlled Open` Story)

### Usage Example

A controlled open state for an Autocomplete that is opened by default can be demonstrated as follows:

```javascript
const [open, setOpen] = React.useState(true);

<Autocomplete
  // Handles Autocomplete close
  onClickAway={() => {
    setOpen(false);
  }}
  
  // Handles Autocomplete Open
  onClick={() => {
    if (open) {
      setOpen(false);
    } else {
      setOpen(true);
    }
  }}
  
  // The controlled open state
  open={open}
  
  // ... Rest of the Autocomplete Props
/>
```

### DIFF
--- a/packages/components/src/core/Autocomplete/components/AutocompleteBase/index.tsx
+++ b/packages/components/src/core/Autocomplete/components/AutocompleteBase/index.tsx
@@ -1,6 +1,7 @@
 import {
   AutocompleteChangeDetails,
   AutocompleteChangeReason,
+  AutocompleteCloseReason,
   AutocompleteFreeSoloValueMapping,
   AutocompleteInputChangeReason,
   AutocompleteRenderInputParams,
@@ -74,6 +75,11 @@ interface ExtraAutocompleteProps<T, Multiple, DisableClearable, FreeSolo>
   ) => void;
   InputBaseProps?: Partial<InputSearchProps>;
   label?: string;
+  onClickAway?: (
+    event?: MouseEvent | TouchEvent,
+    reason?: AutocompleteCloseReason
+  ) => void;
+  onClick?: (event?: TouchEvent | MouseEvent) => void;
 }
 
 type CustomAutocompleteProps<
@@ -119,6 +125,10 @@ const AutocompleteBase = <
     search = false,
     clearOnBlur = false,
     blurOnSelect = !multiple,
+    onClickAway,
+    onClick,
+    onOpen,
+    onClose,
   } = props;
   const theme: SDSTheme = useTheme();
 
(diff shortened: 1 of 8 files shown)


packages/components/src/core/Autocomplete/components/AutocompleteBase/index.tsx
```diff
@@ -141,22 +151,35 @@ const AutocompleteBase = <
       inputValue={inputValue}
       renderInput={defaultRenderInput}
       multiple={multiple}
+      /**
+       * (masoudmanson): Overrides the onOpen and onClose handlers to ensure that
+       * the onClick and onClickAway handlers are triggered when a controlled dropdown
+       * is opened or closed, respectively.
+       */
+      onOpen={defaultOnOpen}
+      onClose={defaultOnClose}
```
tihuan, line 160: Thanks for the documentation!

packages/components/src/core/Autocomplete/components/AutocompleteBase/index.tsx
```diff
@@ -141,22 +151,35 @@ const AutocompleteBase = <
...
-      // (masoudmanson): For multiple options (multiple = true), setting blurOnSelect to false
-      // keeps the dropdown open after selecting an option. This feature enhances user experience
-      // by preventing the dropdown from closing, enabling users to select multiple options
-      // without having to reopen the menu repeatedly.
+      /**
+       * (masoudmanson): For multiple options (multiple = true), setting blurOnSelect to false
+       * keeps the dropdown open after selecting an option. This feature enhances user experience
+       * by preventing the dropdown from closing, enabling users to select multiple options
+       * without having to reopen the menu repeatedly.
+       */
```
tihuan, line 167: thanks for using JSDocs style 🙏 ✨

packages/components/src/core/Autocomplete/components/AutocompleteMultiColumn/index.tsx
```diff
@@ -219,6 +243,21 @@ const AutocompleteMultiColumn = <
...
+             * to handle the closing of the dropdown.
+             */
+            onKeyDown: (event: React.KeyboardEvent) => {
+              if (event.key === "Escape") {
+                if (isOpenControlled) {
+                  onClick?.();
+                  return;
+                }
+
+                setPopperOpen(false);
```
tihuan, line 258: Oh I wonder if we should make it consistent with other event handlers in the sense that we call `props.onKeyDown()` instead of `props.onClick?.()`, so if the component is controlled, the parent will need to handle the onKeyDown event on their own and decide if they want to set open to false and pass it down to the component?

tihuan (commented):
Looks super good, @masoudmanson ! Thanks for the implementation and story update ✨ !!

packages/components/src/core/Autocomplete/components/AutocompleteMultiColumn/index.tsx
masoudmanson: Oh, of course, that makes sense, I'll fix it. :D

tihuan: Thank you!!

tihuan (approved):
Will LGTM first, since I have back to back meetings coming up lol Thanks, Masoud!!

PR #774, opened 2024-05-10
fix(dropdown): fix props propagating

masoudmanson:
Props are now accurately propagated to the underlying button component.


atarashansky (approved):
Approving for what it's worth <3

Issue #823, opened 2024-08-08
Fix ComplexFilter re-rendering bug


masoudmanson:
Reported by: @codemonkey800 

For some reason when changing the options, it re-renders two times for some reason. because of how our application is built, the extra re-render causes a 2nd request to the backend. I'm currently looking into it right now but curious if y'all have any ideas on what could be the issue. more context can be found here:

https://github.com/chanzuckerberg/cryoet-data-portal/issues/747#issuecomment-2264030209

the following sample code was used to test the filter:

https://github.com/chanzuckerberg/cryoet-data-portal/issues/747#issuecomment-2264035698

prior to the upgrade, the above code worked as expected and only re-renders once. however, after the upgrade it now re-renders twice

PR #824, opened 2024-08-08
fix(complexfilter): fix re-rendering bug

masoudmanson:
## Summary

**ComplexFilter**
Github issue: #823

### DIFF
--- a/packages/components/src/core/ComplexFilter/index.tsx
+++ b/packages/components/src/core/ComplexFilter/index.tsx
@@ -176,23 +176,15 @@ const ComplexFilter = <
       if (isTriggerChangeOnOptionClick) {
         setPendingValue(newValue);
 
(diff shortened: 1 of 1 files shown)


packages/components/src/core/ComplexFilter/index.tsx
```diff
@@ -176,25 +176,17 @@ const ComplexFilter = <
...
       return setPendingValue(newValue);
     }
 
-    setValueAndCallOnChange(event, newValue);
+    setValue(newValue);
 
     if (!multiple) setOpen(false);
   }
 
-  function setValueAndCallOnChange(
```
masoudmanson, line 190: We are calling the onChange in a useEffect at line 76, so there is no need for this function anymore!

tihuan (approved):

Issue #827, opened 2024-08-09
Complex filter infinite re-rendering on initial load


codemonkey800:
**Describe the bug**

Initial loading of the data portal results in an infinite re-render bug with the latest version of SDS

**To Reproduce**
1. [Run the data portal in dev mode](https://github.com/chanzuckerberg/cryoet-data-portal/wiki/Quick-Start)
2. Update SDS to `^20.7.0`
3. Go to http://localhost:8080/browse-data/datasets?files=tilt-series
4. See page filter glitching because of infinite re-render

**Expected behavior**

It should not re-render infinitely

**Screenshots**

https://github.com/user-attachments/assets/4faafbbc-d1be-44c6-a6cc-e45534187fa4

PR #828, opened 2024-08-09
fix(complexfilter): fixes infinite re-render issue for SSR hydration

codemonkey800:
## Summary

Github issue: [#827](https://github.com/chanzuckerberg/sci-components/issues/827)
Initial loading of the data portal results in an infinite re-render bug with the latest version of SDS

Fixes issue with the complex filter causing infinite re-rendering on the data portal on initial load.

### Before

https://github.com/user-attachments/assets/5d5ede70-62db-4ef3-8c37-acdc6f08e1c8

### After

https://github.com/user-attachments/assets/d01caa1b-7ee3-44c6-a0d3-33382ded2589

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -3,6 +3,20 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [20.7.0](https://github.com/chanzuckerberg/sci-components/compare/@czi-sds/components@20.1.2...@czi-sds/components@20.7.0) (2024-08-08)
+
+### Bug Fixes
+
+- **a11y:** fix a11y issues ([#787](https://github.com/chanzuckerberg/sci-components/issues/787)) ([e337899](https://github.com/chanzuckerberg/sci-components/commit/e3378997ec2aaf8c4a2d765870668b86461be01c))
+- **complexfilter:** fix prop propagation ([7530284](https://github.com/chanzuckerberg/sci-components/commit/7530284831aec9d62c67b938303eb7f88f775158))
+- **complexfilter:** fix re-rendering bug ([#824](https://github.com/chanzuckerberg/sci-components/issues/824)) ([0692409](https://github.com/chanzuckerberg/sci-components/commit/06924094808ad05e3888507f634fb6c293df007c))
+- **dropdown:** fix props propagating ([#774](https://github.com/chanzuckerberg/sci-components/issues/774)) ([c42970a](https://github.com/chanzuckerberg/sci-components/commit/c42970a2270c5c6c1dda40422cb7eeee9b4b2f5e))
+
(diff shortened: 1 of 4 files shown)


packages/components/src/core/ComplexFilter/index.tsx
```diff
@@ -69,19 +69,23 @@ const ComplexFilter = <
...
-    setPendingValue(value);
-  }, [onChange, value]);
+  const setValue = useCallback(
+    (nextValue: AutocompleteValue<T, Multiple, DisableClearable, FreeSolo>) => {
+      setValueState(nextValue);
+      onChange(value);
+      setPendingValue(nextValue);
+    },
+    [onChange, value]
+  );
```
tihuan, line 82: I also wonder we should compare the existing value in state before calling onChange too, in case the component user passes a new arrow function at each render and the value actually didn't change?

E.g.,

```tsx
<ComplexFilter onChange={() => { ... }} />
```

Something like:

```tsx
const prevValueRef = useRef();

  useEffect(() => {
    if (prevValueRef.current !== value) {
      onChange(value);
      setPendingValue(value);
      prevValueRef.current = value;
    }
  }, [value, onChange]);
```

packages/components/src/core/ComplexFilter/index.tsx
```diff
@@ -69,19 +69,23 @@ const ComplexFilter = <
...
 
-  const [value, setValue] = useState(getInitialValue());
+  const [value, setValueState] = useState(getInitialValue());
   const [pendingValue, setPendingValue] = useState(getInitialValue());
 
-  useEffect(() => {
-    onChange(value);
-    setPendingValue(value);
-  }, [onChange, value]);
+  const setValue = useCallback(
```
tihuan, line 75: hmm it feels like `setValue()` will still get re-created when `onChange` is a new arrow function at reach render. So I'm unsure how this solution is different from the existing one. But I'm probably missing something obvious 😆

tihuan (commented):
Thanks so much for the PR, @codemonkey800 !! Really appreciate the contribution 🎉 🙏 !! Just some quick questions PTAL thank you!

packages/components/src/core/ComplexFilter/index.tsx
tihuan: Ahh great discovery! Yeah it seems like we need to update `ComplexFilter.getInitialValue()` to pass prop.value if available 💡 

Thanks so much for finding the root cause 👏 !!

tihuan (approved):
Looks super good! Thanks so much, @codemonkey800 ! 
Please still wait for @masoudmanson 's review in case I missed something!

masoudmanson (approved):
Thank you so much for taking the time and fixing this bug @codemonkey800 
it LGTM 🥳🎉🤩✨

PR #825, opened 2024-08-08
fix: implement add use client to all sds components

masoudmanson:
## Summary

**All SDS Components**
Github issue: #814 

We nee to add "use client"; in all SDS component files, so projects don’t have to do that manually themselves like here:
https://github.com/czbiohub-sf/CZBSF-Metadata-Tracker-refactor/pull/82#discussion_r1684929746

Issue #814, opened 2024-07-22
Implement: Add "use client"; to all SDS components


masoudmanson:
We nee to add "use client"; in all SDS component files, so projects don’t have to do that manually themselves like here:
https://github.com/czbiohub-sf/CZBSF-Metadata-Tracker-refactor/pull/82#discussion_r1684929746

PR #831, opened 2024-08-21
fix(rollup): add "use client;" banner on top of rollup outputs files

masoudmanson:
## Summary

Github issue: #814 
Old PR: #825 

We need to add "use client"; in all SDS component files so that projects don’t have to do it manually, as shown here:
[Example PR Discussion](https://github.com/czbiohub-sf/CZBSF-Metadata-Tracker-refactor/pull/82#discussion_r1684929746).

Thanks to @tihuan (https://github.com/chanzuckerberg/sci-components/pull/825#pullrequestreview-2228530236), we now know that setting a banner in the Rollup config will automatically add the "use client"; directive to the top of output files.

Since we don’t use preserveModules: true in our Rollup configs, we don’t need to use the [rollup-plugin-preserve-directives](https://github.com/Ephem/rollup-plugin-preserve-directives), which would otherwise add the directive to every single module file!

Refrences:
https://github.com/rollup/rollup/issues/4699#issuecomment-2273497055
https://github.com/rollup/rollup/issues/4699#issuecomment-2275883910

### DIFF
--- a/packages/components/rollup.config.mjs
+++ b/packages/components/rollup.config.mjs
@@ -26,10 +26,12 @@ const config = [
     input: "src/index.ts",
     output: [
       {
+        banner: "'use client';",
         file: pkg.main,
         format: "cjs",
       },
       {
+        banner: "'use client';",
         file: pkg.module,
         format: "esm",
       },


tihuan:
CC: [kyleawayan](https://github.com/kyleawayan) - "use client" is being added to component files in this PR 🎉 It should be released next Wednesday 🕺 Thanks for waiting!

Issue #625, opened 2023-10-19
`defaultExpanded` MUI prop doesn't work with SDS `Accordion` component


liaprins:
**Describe the bug**
The SDS `Accordion` component is built off the MUI `Accordion` component, which has a prop of `defaultExpanded`. When `true` / present, this boolean prop forces the Accordion to be open by default (users can still close it). However, this does not work with SDS `Accordion` if this prop is manually added; it has no effect.

**To Reproduce**
1. Go to the index.stories.tsx file for the `Accordion` component and open Storybook
2. For one of the `Accordion` components, add the `defaultExpanded` prop
3. Check that component in Storybook
4. It will not be expanded by default

**Expected behavior**
The Accordion with the `defaultExpanded` prop should have been expanded

**Additional context**
Once this is fixed, update the `Accordion`'s Zeroheight documentation to remove the bullet about eh `defaultExpanded` prop not working, form the MUI vs SDS section.

PR #838, opened 2024-08-28
fix(accordion): defaultExpanded now works as expected

masoudmanson:
## Summary

**Accordion**
Github issue: #625 

The SDS Accordion component is built off the MUI Accordion component, which has a prop of defaultExpanded. When true / present, this boolean prop forces the Accordion to be open by default (users can still close it). However, this does not work with SDS Accordion if this prop is manually added; it has no effect.

The Accordion with the defaultExpanded prop should have been expanded!


tihuan (approved):
Oh perfect!! Thanks for adding support for this prop, @masoudmanson 😄 🙏 LGTM!!

clarsen-czi (changes requested):
@masoudmanson looks like this isn't working in Storybook via Chromatic. Not sure if there's actually an issue or if Chromatic is just being fussy. Take a look and LMK. Thanks!

masoudmanson:
@clarsen-czi, could you please check this link:
https://616981d9028812003ade73f3-pciozcnerq.chromatic.com/?path=/story/components-accordion--default&args=defaultExpanded:!true

After enabling the defaultExpanded prop, you'll need to refresh the page to see the Accordion expanded by default. The link above already has the prop set, so you should see the Accordion in its expanded state. Let me know if it's still not working!

clarsen-czi:
> @clarsen-czi, could you please check this link: https://616981d9028812003ade73f3-pciozcnerq.chromatic.com/?path=/story/components-accordion--default&args=defaultExpanded:!true
> 
> After enabling the defaultExpanded prop, you'll need to refresh the page to see the Accordion expanded by default. The link above already has the prop set, so you should see the Accordion in its expanded state. Let me know if it's still not working!

Looks like it's working! Thanks for providing the link to check @masoudmanson

clarsen-czi (approved):
Looks like it's working!! Thanks @masoudmanson

Issue #849, opened 2024-09-19
Add `success` intent on `InputText`


masoudmanson:
The `InputText` component currently supports `notice`, `negative`, and `default` intents. We need to add a `success` intent to accommodate future form validation components.

PR #857, opened 2024-09-26
feat(inputs): add positive intent to input compoennts

masoudmanson:
## Summary

**Inputs**
Github issue: #849 

The `InputText` component and rest of the input components currently supports `notice`, `negative`, and `default` intents. We need to add a `positive` intent to accommodate future form validation components.

### DIFF
--- a/packages/components/src/core/InputDropdown/style.ts
+++ b/packages/components/src/core/InputDropdown/style.ts
@@ -25,12 +25,14 @@ const doNotForwardProps = [
   "sdsStyle",
 ];
 
+type IntentType = "negative" | "notice" | "positive";
+
 export interface InputDropdownProps
   extends CommonThemeProps,
     Omit<ButtonProps, (typeof doNotForwardProps)[number]> {
   children?: ReactNode;
   disabled?: boolean;
-  intent?: "default" | "negative" | "notice";
+  intent?: "default" | IntentType;
   label: ReactNode;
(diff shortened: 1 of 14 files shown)


packages/components/src/core/InputDropdown/style.ts
```diff
@@ -25,12 +25,14 @@ const doNotForwardProps = [
   "sdsStyle",
 ];
 
+type IntentType = "negative" | "notice" | "positive";
```
tihuan, line 28: just curious, should "default" be part of the IntentType?

tihuan (commented):
Looks v good thank you!! Just one tiny comment otherwise LGTM 😁 🥇

packages/components/src/core/InputDropdown/style.ts
masoudmanson: Hey @tihuan ,

Thanks for the review! Since our theme doesn’t include a `"default"` color intent, I had to keep them separate. However, the `intent` prop includes `"default" | IntentType` to cover all possible values. 🫠

clarsen-czi (approved):
LGTM!! Thanks @masoudmanson!

packages/components/src/core/InputDropdown/style.ts
tihuan: OH!! That's why 😆 TIL thank you!

Issue #816, opened 2024-08-05
Design: `NavigationHeader`


clarsen-czi:
**Background:**
A Header Navigation component is a high priority component to be added into SDS for CZ One Science teams to use in their project work. `NavigationHeader` component was started as part of SDS Hack Day in Q4 2023 but the project was never completed. Biohub UI, CryoET, and Cell Science all have Header Navigation components implemented that can be referenced when designing and building an SDS version.

PR #885, opened 2024-10-30
feat(NavigationHeader): implement NavigationHeader component

codemonkey800:
## Summary

Github issue: #816

Implements the NavigationHeader component for desktop and mobile

## Checklist

- [x] Test Story in Storybook
- [x] Tests written


kev-zunshiwang:
Looking great, thanks Jeremy!!

Some feedback below:
- Logo Slot
  - Update color to Base/border for the dash border line around "logo slot"
  - Unbold the text "logo slot"
- Tag
  - Update text inside the tags to title case
  - Remove the hover effect for the tags
- Buttons
  - Update the "person" minimal button to a icon button (see design for the specific spec)
- The entire area of "logo slot / logo name / beta tag" should be linked to the home page of the product. Not sure if this affects the code

kev-zunshiwang:
Mobile Feedback
Logo
- whole area should be a link

Secondary Nav Item
- the whole row should be a target area - see spec for details
- On hover, the background color of the whole row should have a gray (base/surface-secondary)
Secondary Nav Sub Item
- the whole row should be a target area - see spec for details
- On hover, bold the text with Narrow/fontBodyS/600 in addition to the background color change.

Buttons
- the primary / secondary buttons / minimal button should extend to the whole row - see spec for detals
- Use the S size of the person icon, the illustration is different from the larger size
- bold the text - Narrow/fontBodyM/600

packages/components/src/core/Button/index.tsx
```diff
@@ -12,28 +12,31 @@ import { filterProps } from "src/common/utils";
...
+}
+
+export interface SdsButtonProps {
+  sdsStyle?: "rounded" | "square";
+  sdsType?: "primary" | "secondary" | "destructive";
+  isAllCaps?: boolean;
+  isRounded?: boolean;
+}
+
+type SdsProps = SdsIconButtonProps | SdsMinimalButtonProps | SdsButtonProps;
```
tihuan, line 39: thanks for cleaning this up!! 🙏 ✨

tihuan (approved):
LGTM!! 

Not a blocker, but IIRC many SDS components are wrapped in `forwardRef()` not sure if it makes sense to do it here too as a followup PR 😄 

Thanks again!!

clarsen-czi (changes requested):
Looking good @codemonkey800! There are some changes to make so please reach out if you have any questions about the comments I've added. Thanks!!

tihuan:
> LGTM!!
> 
> Not a blocker, but IIRC many SDS components are wrapped in `forwardRef()` not sure if it makes sense to do it here too as a followup PR 😄
> 
> Thanks again!!

Actually React 19 is dropping ForwardRef! But we probably still need to add forwardRef until all of our projects upgrade to React 19 😁 

https://react.dev/blog/2024/12/05/react-19

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onChange,
+  value,
+}: NavigationHeaderPrimaryNavProps<T>) {
+  return (
+    <StyledSection>
+      {items.map((item) => {
+        const isActive = item.key === value;
+
+        return (
+          <PrimaryNavItem
```
tihuan, line 33: not a blocker for this PR, but is there a way to accept ReactNode or somehow allow rendering Next/Link and/or `<a />` for links instead of onClick driven navigation? Just thinking that would help with semantic HTML + a11y!

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export interface NavigationHeaderPrimaryNavProps<T extends string> {
+  hasInvertedStyle?: boolean;
+  items: NavigationHeaderPrimaryNavItem<T>[];
+  onChange(value: T): void;
+  value: T;
+}
+
+export default function NavigationHeaderPrimaryNav<T extends string>({
```
tihuan, line 21: This could be VCP's design not adhering to SDS recommendations, but VCP actually has a primary navigation with dropdown. Should that be a supported use case?

Screenshot 2024-12-05 at 2 22 07 PM


CC: @clarsen-czi 🙏

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/style.ts
```diff
@@ -0,0 +1,65 @@
...
+    background: none;
+  }
+
+  ${(props: PrimaryNavItemProps) => {
+    const spaces = getSpaces(props);
+    const colors = getSemanticColors(props);
+
+    return `
+      gap: ${spaces?.xs}px;
+
```
tihuan, line 25: extra line 😁 
```suggestion
```

packages/components/src/core/NavigationHeader/components/NavigationHeaderSecondaryNav/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+          </Menu>
+        </>
+      )}
+
+      {item.type === "dropdown" && isNarrow && (
+        <StyledAccordion
+          id={`${item.label}-dropdown`}
+          hasInvertedStyle={hasInverseStyle}
+        >
+          <AccordionHeader>{item.label}</AccordionHeader>
```
tihuan, line 106: such a great idea of using Accordion here 🔥

tihuan (commented):
Some extra comments after attempting to use it for VCP TopNav! I ended up cherrypicking the code from this PR for the VCP implementation due to extreme time constraint. But look forward to replacing the temp solution with SDS NavigationHeader in 2025 🎉 !!

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
tihuan: Got it! Yeah only if design sees primary nav dropdown as a common pattern instead of a one time snowflake 😁 👌

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
tihuan: In other components we export the subcomponents to allow the users to compose to maintain the flexibility and retain the styling consistency. In this case, perhaps exporting `PrimaryNavItem` could help? That way the Nav Header can accept ReactNode and the user can wrap `PrimaryNavItem` with a Link component if needed? 

Something like that, probably still need to think through together 😂

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
clarsen-czi: I believe Kevin's intention was that primary navigation would not include dropdowns and that dropdowns are meant to only be available on secondary nav items. It seems to me that, in the example screenshot above, the Tutorials section actually makes more sense as secondary navigation since it links to training/how-to content as opposed to core functional area of application. @kev-zunshiwang let us know what you think!

tihuan: Ahh I see! So more like VCP not following design best practices 💡 That makes sense to me 😁 Thank you!

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
masoudmanson: @codemonkey800, I agree with having a `LinkComponent`, it makes total sense. I’ve also been thinking about refactoring some of our other components to use a polymorphic architecture for better control over the underlying tags. 😬🥸

masoudmanson (approved):
It looks great to me, @codemonkey800 ! 

Thank you so much for your help on this component, it really means a lot to us. 
I just made some final tweaks to the colors and double-checked everything with Connor.

packages/components/src/core/Button/index.tsx
```diff
@@ -112,6 +115,7 @@ const Button = React.forwardRef(
             variant="text"
             {...propsWithDefault}
             ref={ref}
+            data-style={sdsStyle}
```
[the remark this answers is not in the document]
tihuan: Ahh brilliant! Thanks for sharing your learning, @masoudmanson 🎉 Adopting this approach going forward sounds awesome!

packages/components/src/core/NavigationHeader/components/NavigationHeaderPrimaryNav/index.tsx
tihuan: Oh should we add `LinkComponent` support for `PrimaryNavItem` and `SecondaryNavItem` 😄 🙏 ?

Thank you!!

clarsen-czi (changes requested):
Looking really good @masoudmanson ! Just a few changes to be made mostly on the mobile view and we should be good to go!

clarsen-czi (changes requested):
Looks great @masoudmanson! Everything we discussed you've gotten fixed.

The only reason why I'm saying "request changes" is I think we should change when the mobile view takes over: if the viewport isn't wide enough to fit the whole desktop version of the nav in it, then it should switch to the mobile version as opposed to becoming left/right scrollable as it currently does. Scrollable nav bars (and footers for that matter) are not great UX.

clarsen-czi (approved):
LGTM!! Thanks so much for all the hard work on this component @masoudmanson !! This was a huge lift and a ton of work 💪

PR #919, opened 2024-12-18
fix(ButtonDropdown): Sync Button props type

crispipear:
We noticed that the prop "minimal" is not recognized in Typescript on the ButtonDropdown, which trace down to the missing "minimal" string to the ButtonDropdown's `sdsStyle`.

Since `sdsStyle` and `sdsType` are used directly into <Button />, I think it would be more type-safe to have these corresponding props in ButtonDropdown to use the exact same type as Button.

## Summary

**Structural Element (Base, Gene, DNA, Chromosome or Cell)** Base

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written


masoudmanson (approved):
Thank you so much, @crispipear! 🙏🥳
The `Button` and `ButtonDropwdown` types have changed recently, so the old ones were a bit outdated. I updated them using your logic and adapted it to the new types.

Really appreciate your contributions, this PR looks great to me! ✨✨
I’ll make sure both of your PRs are included in the next release, most likely on Wednesday, April 2nd.

masoudmanson:
@crispipear 

You also mentioned that the `minimal` variant isn’t available in the TypeScript types for `ButtonDropdown`. That’s because we don’t currently have a `minimal` version of `ButtonDropdown`, the closest alternative would be `InputDropdown`, which does support a `minimal` variant.

Here is an example: [SDS minimal InputDropdown](https://chanzuckerberg.github.io/sci-components/?path=/story/components-inputs-inputdropdown--default&args=sdsStyle:minimal)

I’ve reached out to our design team to confirm whether we want to introduce a `minimal` variant for `ButtonDropdown`. If they’re on board, I’ll go ahead and add it, and I’ll let you know once it’s released and ready to use.

Issue #927, opened 2025-01-28
Add `hover` prop to the `HeaderCell` component


masoudmanson:
- [x] Implementation
- [ ] Documentation
- [ ] Test

PR #928, opened 2025-01-28
fix(headercell): added new hover prop to control component hover state

masoudmanson:
## Summary

**HeaderCell**
Github issue: #927 

https://czi-sci.slack.com/archives/C032S43KKFV/p1738019694310679

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/src/core/CellHeader/style.ts
+++ b/packages/components/src/core/CellHeader/style.ts
@@ -12,6 +12,7 @@ export interface CellHeaderExtraProps extends CommonThemeProps {
   active?: boolean;
   hideSortIcon?: boolean;
   horizontalAlign?: "left" | "center" | "right";
+  hover?: boolean;
 }
 
 const contentPositionMapping = {
@@ -27,6 +28,7 @@ const doNotForwardProps = [
   "tooltipProps",
   "tooltipText",
   "hideSortIcon",
+  "hover",
 ];
 
 export const StyledSortingIcon = styled(Icon, {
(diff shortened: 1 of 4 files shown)


packages/components/src/core/CellHeader/style.ts
```diff
@@ -55,28 +57,36 @@ export const StyledTableHeader = styled("th", {
   ${focusVisibleA11yStyle}
 
   ${(props: CellHeaderExtraProps) => {
-    const { active = false, horizontalAlign = "left" } = props;
+    const { active = false, horizontalAlign = "left", hover = true } = props;
```
ehoops-cz, line 60: It looks like it defaults to "true" 👍

ehoops-cz (approved):

packages/components/src/core/CellHeader/style.ts
masoudmanson: Is it ok? or should it be the other way around? 🤔
What do you think @clarsen-czi?

clarsen-czi (approved):
LGTM!! Thanks for jumping on this so quickly @masoudmanson!

packages/components/src/core/CellHeader/style.ts
tihuan: @clarsen-czi sorry I think we still need your feedback here 😁 🙏 

If we default `hover` to `true`, it would mean that the `CellHeader` could have hover effect when no corresponding action is present/implemented, so the question here is whether we should default `hover` to `false` instead to prevent that?

Thank you!

clarsen-czi: Good call @tihuan! I agree that having it default to `false` makes the most sense. Thanks for the suggestion!

PR #939, opened 2025-03-14
fix(ContentCard): Fixed ContentCard content height + Fixed TS types on CC

masoudmanson:
## Summary

**Content Card**
Jira issue: [SCIDS-984](https://czi.atlassian.net/browse/SCIDS-984)

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

[SCIDS-984]: https://czi.atlassian.net/browse/SCIDS-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/core/Button/index.tsx
+++ b/packages/components/src/core/Button/index.tsx
@@ -31,12 +31,14 @@ export interface SdsIconButtonProps extends BaseButtonProps {
   sdsSize?: ButtonSize;
   icon: keyof IconNameToSizes | React.ReactElement<CustomSVGProps>;
   sdsIconProps?: Partial<IconProps<keyof IconNameToSizes>>;
+  target?: "_blank" | "_self" | "_parent" | "_top" | "framename";
 }
 
 export interface SdsMinimalButtonProps extends BaseButtonProps {
   sdsStyle: "minimal";
   sdsType?: ButtonTypeMap["minimal"];
   children: React.ReactNode;
+  target?: "_blank" | "_self" | "_parent" | "_top" | "framename";
 }
 
 export interface SdsButtonProps extends BaseButtonProps {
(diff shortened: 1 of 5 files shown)


packages/components/src/core/Button/index.tsx
```diff
@@ -23,6 +23,7 @@ type ButtonSize = "small" | "medium" | "large";
 interface BaseButtonProps extends RawButtonProps {
   children?: React.ReactNode;
   isAllCaps?: boolean;
+  target?: "_blank" | "_self" | "_parent" | "_top" | "framename";
```
tihuan, line 26: discussion: instead of hardcoding this, I wonder we should use an existing type? Like maybe `AnchorHTMLAttributes<HTMLAnchorElement>["target"];` 😄 ?

packages/components/src/core/ContentCard/index.tsx
```diff
@@ -205,12 +205,12 @@ const ContentCard = React.forwardRef<HTMLDivElement, ContentCardProps>(
       >
         {clickableCard ? (
           <StyledCardActionArea
+            {...clickableCardProps}
```
tihuan, line 208: can we add a comment why this spread needs to happen at the top instead of being at the bottom, so the user can overwrite the default props please 🙏 ? Thank you!

tihuan (approved):
Looking great!! Thanks so much, Masoud 😄 Just two non blocking tiny questions thank you!!

packages/components/src/core/Button/index.tsx
masoudmanson: Oh, sure! That makes perfect sense. 🥸

packages/components/src/core/ContentCard/index.tsx
masoudmanson: Absolutely 🙃🙂

PR #980, opened 2025-07-29
fix(navigationheader): fixed dark mode and inverted mode styles

masoudmanson:
## Summary

**NavigationHeader**

Jira issue: https://czi.atlassian.net/browse/SCIDS-1205?atlOrigin=eyJpIjoiNjkyMWEzOGNkY2ExNDk3ZDlkMmNhZDRlNDY5YzRmNTAiLCJwIjoiaiJ9

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written

### DIFF
--- a/packages/components/src/core/NavigationHeader/style.ts
+++ b/packages/components/src/core/NavigationHeader/style.ts
@@ -495,6 +495,7 @@ export const StyledButtonSection = styled("section", {
 })`
   display: flex;
   align-items: center;
+  z-index: 10;
 
   .MuiButtonBase-root {
     ${fontBody("l", "semibold", /* isNarrow */ true)}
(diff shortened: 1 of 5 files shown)


packages/components/src/core/NavigationHeader/style.ts
```diff
@@ -495,6 +495,7 @@ export const StyledButtonSection = styled("section", {
 })`
   display: flex;
   align-items: center;
+  z-index: 10;
```
rainandbare, line 498: What happened here?

rainandbare (approved):
Looks great! Thank you Masoud! 🥹

packages/components/src/core/NavigationHeader/style.ts
masoudmanson: The new section titles had a none zero `z-index`, which caused them to cover the sticky buttons on mobile when scrolling. I’ve updated the `z-index` so the sticky buttons stay on top and the issue is fixed. 😏🥳

PR #991, opened 2025-08-18
fix(contentcard): fixed Clickable ContentCard hover styles

masoudmanson:
## Summary

**ContentCard**

Fix hover styles for the `ContentCard` component


tihuan (approved):
Awesome thanks for the fix 🚀!

PR #1032, opened 2025-11-05
fix(navigationheader): fix navigation header action buttons style in dark mode

masoudmanson:


clarsen-czi (approved):
LGTM!! Thanks @masoudmanson !

PR #1034, opened 2025-11-05
feat(navigationhedaer): optional header and footer titles

masoudmanson:


clarsen-czi (approved):
LGMT! Thanks @masoudmanson!!

PR #1036, opened 2025-11-05
fix(navigationhedaer): nav header title

masoudmanson:


clarsen-czi (approved):
Thanks @masoudmanson !

PR #1109, opened 2026-03-09
feat: add Emotion babel plugin to Rollup build (Vite support)

kyleawayan:
Adds `@emotion/babel-plugin` to the components Rollup build so the published dist includes `__emotion_base` on styled components. This is required for Emotion component selectors (`${StyledChild} { ... }`) to work in bundlers that don't run the Emotion plugin themselves (e.g. Vite/esbuild).

Next.js works out of the box because it has its own handling with Emotion which takes care of this by default.

Without this, consumers get a "Component selectors can only be used in conjunction with @emotion/babel-plugin" error.

- [x] Tested with Vite + TanStack Start
- [x] Tested with `create-sds-app`

Related Slack thread: https://czbiohubnetwork.slack.com/archives/C032S43KKFV/p1772829692628259?thread_ts=1772828928.465179&cid=C032S43KKFV


tihuan (approved):
LGTM! Please wait for Masoud's review before merge thank you 🙏

masoudmanson (approved):
Thank you so much @kyleawayan LGTM :D

kyleawayan:
Ok sorry this is going deeper than it seems. Finding that both in my TanStack Start app and the `create-sds-app` test I did... my local installation of SDS does not have `__emotion_base` in dist..., yet I did something in between to get it to work with TanStack Start.. sorry all! Will switch to draft PR while I take a look at this!

kyleawayan:
Ok all fixed! Please review if needed again @masoudmanson and @tihuan -- the reason why I got this:

> Ok sorry this is going deeper than it seems. Finding that both in my TanStack Start app and the create-sds-app test I did... my local installation of SDS does not have __emotion_base in dist..., yet I did something in between to get it to work with TanStack Start.. sorry all! Will switch to draft PR while I take a look at this!

Was because of the order of the babel plugin. Turns out I had it compiling nothing haha.

Tested in fresh version of Astro and TanStack Start. Only thing I noticed in testing is that the Dropdown is broken (`"Cannot read properties of undefined (reading
  'ref')"`) -- Claude says it may be a React 19 issue (because of the fresh Astro/TanStack Start templates) but this is out of scope of this PR

masoudmanson (approved):
It LGTM, @kyleawayan! 🤩🥳 Thank you so much for testing this thoroughly and covering the Astro setup as well.

PR #1145, opened 2026-07-16
feat(data-viz): Improve stacked bar chart animation logic

bchu1:
## Summary

Fixes 2 issues:

- Makes bars transition in 1 instead of 2 animation sequences.
- Fixes stutter when more bars are getting added.

Before:

https://github.com/user-attachments/assets/46c62596-bd9b-4035-adc9-cc03c85492c3

After:

https://github.com/user-attachments/assets/ebd232d0-2c3e-4cee-aadc-bb6ca680ddbb

The PR accomplishes this by reworking the animation sequences into 3 steps:

1. Keep existing bars, but add new bars with 0 width (no visible change).
2. Update all bars to new width, with removed bars shrinking to 0 width but still in the DOM (this performs the animation).
3. Remove removed bars by updating state to final array (no visible change, cleanup only).

Notes:

- Changed Storybook data to be selector with different arrays
- Added `isEntering` and `isExiting` properties to segments for when they should be rendered with 0 width for transition purposes
- Simplified effect logic, no longer need separate function definition
- `setAnimatedRemainingPercentage()` is only called in step 2, along with the other transitions.
- `previousProcessedRef` must only be updated in the final step (3), so that if the effect triggers again before all steps are complete, it does not use the new value before it has already finished rendering.
- Unused `index` param to `renderBarSegment()` removed
- The remaining bar now always exists in the DOM, even if it is width 0 (this is the easiest way to support transitions for it. Ideally, we would compress that into the other data array so that it is not special, but that would be a bigger diff right now). It, however, does not get painted on initial render, like the other bars which get painted only after the `useEffect()` has run.


tihuan (commented):
Ah nice improvement and code comments ✨ Thank you so much!

Maybe wait for Masoud to review, since he has more context 😁

masoudmanson:
@bchu1 Thank you so much for the improvement, Bryan! It feels much smoother now, love it! 😍

masoudmanson (approved):

PR #1166, opened 2026-07-29
feat(docs): pull Zeroheight docs into Storybook

masoudmanson:
## Summary

**Structural Element: repo-wide (documentation infrastructure), and a new `@czi-sds/icons` package**

This PR moves the Science Design System documentation from Zeroheight into this repository. Writers now keep the documentation with the code that it describes. Storybook shows the documentation. The build, the CI workflows, and the MCP server no longer use Zeroheight.

This PR also adds `@czi-sds/icons`, a new package for the icons that SDS draws itself, and deprecates the `Icon` component.

### New `@czi-sds/icons` package

SDS now uses [Phosphor Icons](https://phosphoricons.com) for its icon set. The new `@czi-sds/icons` package adds the 16 icons that SDS draws itself. An icon stays in SDS for one of two reasons. Either Phosphor has no equivalent, which covers the brand marks and the concepts that belong to this science. Or Phosphor has something close, and the design team wants a drawing of its own.

The icons are `AtlasIcon`, `BiohubIcon`, `ChartPieFilledIcon`, `ChartPieXIcon`, `CollapseIcon`, `ExpandIcon`, `FoldIcon`, `GenerateIcon`, `GitHubIcon`, `MaximizeSmallIcon`, `MinimizeSmallIcon`, `PredictIcon`, `SparkleIcon`, `SparklesIcon`, `UmapIcon`, and `WandSparkleIcon`.

**One API for both packages.** Each icon is built on Phosphor's `IconBase`, in the way that Phosphor documents for custom icons. An SDS icon therefore takes the same props as a Phosphor icon: `size`, `color`, `weight`, `mirrored`, `alt`, and every prop of an `svg` element. It also reads the same `IconContext`, and thus one provider styles the icons of both packages. The only difference is the import.

```tsx
import { HeartIcon } from "@phosphor-icons/react";
import { AtlasIcon } from "@czi-sds/icons";
```

`@phosphor-icons/react` is a peer dependency and not a bundled one. This keeps one copy of Phosphor in the application, and that single copy is what lets one `IconContext` reach both packages. A test asserts this inheritance, so a bundled copy would fail the suite.

**The artwork.** Each icon is path data on the 256 by 256 grid that `IconBase` renders, as a single `path` or a `Fragment` of them. Design draws at 16 and 32, so the coordinates are scaled once, when the icon is added. The scaling uses exact decimal arithmetic. A rendering test compared every icon before and after the change and found the same pixels. Each icon has one drawing, where Phosphor has six, one per weight. An SDS icon therefore accepts `weight` and looks the same at every value.

**The package.** It does not depend on `@czi-sds/components`. It has no theme, no Emotion, and no Material UI, and thus an application can use the icons without the rest of the design system. It ships ESM and CommonJS with types and an `exports` map, and it has 28 tests. The version is `0.1.0`, because the set of icons may still change.

Two problems in the published Phosphor package shaped this work, and the code says so where it matters. The entry that Phosphor documents for `IconBase`, `@phosphor-icons/react/lib`, has no ESM file in the package, so the import comes from `@phosphor-icons/react/dist/lib/IconBase`. A test fails if that path ever moves. Phosphor also declares itself an ES module and ships its CommonJS build as a `.js` file, so Node reads that build as ESM and it exports nothing. A raw `require` therefore cannot use the icons. Bundlers and `import` are not affected. The Icons overview page records this.

### `Icon` is deprecated

The `Icon` component in `@czi-sds/components` still works, and a later major version will remove it. It now carries a `@deprecated` note that names the replacement, its stories moved to `Deprecated/Icon`, and the page of deprecated components lists it.

There is no runtime warning yet. About 25 SDS components still render `Icon` inside themselves, and a warning would reach a reader who never used the component. That warning belongs with the change that moves those components to the new icons.

### Icon documentation

- **3 pages** in `packages/icons/src/__storybook__/docs/`. The overview has the installation, the props, and the shared defaults. A second page shows the 16 icons and says why each one stays in SDS. A third page maps each prop of the old `Icon` component to its replacement. These pages follow the structure of the overview pages of the other packages, with the links to npm and to GitHub.
- **7 examples**, and 1 more for the card in the catalog.
- One story, `Icons/Overview → Default`, with a control for each prop.
- The catalog on the Overview page now has an `Icons` card. The card for the `Icon` component is gone, because the test in `docs-kit/__tests__/catalog.test.ts` holds the catalog to the stories of the listed sections.
- The design page for icons now describes the design only, and points to the code pages for the API.

### Documentation in Storybook

- **43 design pages** in `design-docs/pages/`. These pages describe the foundations: colors, typography, spacing, corners, borders, drop shadows, icons, theming, and responsive design. They also describe the design part of each component. Storybook supplies 1,074 image files from `/design-assets`.
- **55 component documentation pages**: 52 pages in `components` and 3 pages in `data-viz`. Each page is in the `__storybook__/docs/` folder of its component. Storybook attaches each page to the stories of that component. Therefore the sidebar shows "Documentation" above the stories.
- **229 code examples**. A reader can run each example. Storybook loads each example separately and shows it next to its source code.
- **4 MCP server pages** in `packages/mcp/docs/`. These pages stay in the package that they describe.
- The `storySort` property in `.storybook/preview.jsx` sets the order of the sidebar. To move a page, move one line. The sidebar shows each deprecated component in a different color and puts a line through its name.

### Overview page and component catalog

A new page shows a catalog of all the components in both packages. The catalog puts the components into groups by the task of the reader. The design pages use different groups.

Each card shows the default example of its component as a small copy, and links to the page of that component. The browser makes a card only when the reader scrolls to it. The test in `docs-kit/__tests__/catalog.test.ts` compares the catalog with the libraries. Each component that has stories occurs one time, and each example name refers to a file.

### Live playground

The `playground/` folder adds an editor above an SDS preview. The editor uses Monaco. The playground is a Storybook story, and therefore it gets the library, the theme, and the deployment of Storybook. The playground keeps the code in the URL, and thus a user can send a link to any version of the code. Each documentation example has a link to the playground, and the playground opens with the code of that example.

The command `yarn playground:types` makes the type definitions for the editor. The commands `yarn start` and `yarn build-storybook` run this command automatically. The browser downloads these definitions and does not include them in the bundle. Therefore an absent build cannot break the page.

### New `docs-kit/` folder

The `docs-kit/` folder has the alias `@sds-docs/*`. It contains the components that show all the documentation pages:

- `SdsDoc` shows the documentation HTML with text styles that are correct in the light theme and in the dark theme. It makes a `NavigationJumpTo` sidebar from the headings of the page. It also moves the live previews into the page, and thus the text styles do not change the component styles.
- `SdsExample` loads one example, shows it with the SDS theme, and shows its source code.
- `SdsCatalog`, `CodeFigure`, `highlight`, and `useThemeMode` control the catalog grid, the code blocks, the colors in the code blocks, and the change between the light theme and the dark theme.

The command `npx tsc --noEmit -p tsconfig.docs-check.json` checks the types of each example against the package sources. The package configurations cannot do this check, because the examples import `@czi-sds/components` and `@czi-sds/data-viz`. Vitest tests `docs-kit` and `playground`. The command `yarn test` also runs these tests.

### MCP server

- This PR removes the Zeroheight integration. The files `export-zeroheight-pages.ts`, `zeroheight-config.ts`, `list-styleguides.sh`, `zeroheight-cache.json`, and `.env.example` are no longer in the repository. The API credentials for these files are also no longer necessary.
- The new file `generate-component-docs.ts` makes the MCP documentation from the documentation HTML in this repository.
- The prop extractor now uses the new shared file `lib/props-tables.ts`.
- The resource `sds://rules/zeroheight` has the new name `sds://rules/documentation`.
- The `push-tests`, `namespace-check`, and `release` workflows no longer have `ZEROHEIGHT_API_CLIENT`, `ZEROHEIGHT_API_KEY`, and `ZEROHEIGHT_STYLEGUIDE_ID`.

### Style Dictionary 4 to 5

Style Dictionary version 5 removed the `{token.path.value}` reference. Therefore this PR changes each token reference to `{token.path}`.

SDS keeps the light value and the dark value on one token, and also the wide value and the narrow value. These values have the names `darkValue` and `narrowValue`. SDS also keeps the Inter font stack as a plain text value. The resolver in Style Dictionary cannot follow these references. The new `resolveCustomReferences` preprocessor puts these values into the tokens before the resolver starts. The resolver then resolves and checks the usual token references.

This PR also changes Storybook to version 10.5.6.

### Component corrections

- **Panel**: The SDS theme makes the MUI transitions very short: 20 ms to open and 10 ms to close. A large surface then moves too quickly. Panel now uses the transition times of the MUI drawer. The new `transitionDuration` prop can change these times.
- **NavigationJumpTo**: The component made the tab ids and the React keys from the title of each item. Two items with the same title, for example an "Overview" in each section, then had the same id. The component now adds a number to each title that occurs more than one time.
- **Hero**: The inverted text used the inverse color of the theme. In the dark theme this color is almost black, but the backdrop keeps its color. The text now uses `textPrimaryOnDark`.
- **Tabs**: The component now sends `sdsSize` to the styled tabs element.
- This PR updates the snapshots for these corrections.

The `README.md` file now refers to Storybook, not to Zeroheight.

## Checklist

- [x] Test Story in Storybook
- [x] Tests written
- [x] Semantic Variables from `defaultTheme.ts` used wherever possible
- [x] Documents updated


masoudmanson:
@tihuan @hthomas-czi @clarsen-czi  
When you have a moment, please review the updated Storybook 🥳 😎

I know it’s a huge PR, but it’s mostly documentation and MCP server files. I’ve verified that none of the build processes or Storybook previews break and everything is working as before.

tihuan (approved):
Woohoo! Massive win 🔥 🚀 !! Love the new doc site so much 🤩 👏 Thanks so much, Masoud!!

clarsen-czi (changes requested):
This is looking so so good @masoudmanson !! Really impressed with all the work here. Just a few small tweaks to make and a couple of questions I left that I'm curious to hear your thoughts on.

clarsen-czi (approved):
LGTM!! Just one small update on the Icons table to add in the 10px x 10px size, otherwise we're good to publish. Thanks @masoudmanson !!

PR #1191, opened 2026-09-17
feat(ProteinStructureViewer): add mmcif and pdbx support to mol star

masoudmanson:
# feat(ProteinStructureViewer): accept mmCIF, rename `pdb` to `structure`, and change the chain effect

## Summary

Jira issue: [#XXXX](link)

This pull request makes three changes to the `ProteinStructureViewer` component:

1. The viewer accepts mmCIF (PDBx) text. Before, it accepted only PDB text.
2. The `pdb` prop has a new name. The `format` prop is not necessary. Both changes break the API.
3. The viewer shows a selected chain and a pointed chain in the same way.

### 1. Two structure formats

The `structure` prop accepts PDB text and mmCIF (PDBx) text. The viewer reads the
format from the text:

- Text that has a `data_` block is mmCIF.
- Text that has an `_atom_site.` loop is mmCIF.
- All other text is PDB.

The viewer does not accept binary CIF (`.bcif`), because the prop is a string.

The `plddt` prop works with both formats. For mmCIF text, the viewer writes the
scores into the `_atom_site.B_iso_or_equiv` column. For PDB text, the viewer
writes the scores into columns 60 to 66.

The package exports two new items:

- `detectStructureFormat` reads the format of structure text.
- `StructureFormat` is the type of the result.

### 2. Changes that break the API

The component is new, and no consumer uses it. Therefore these changes are safe.

Do these two steps in the code of a consumer:

1. Change the name of the `pdb` prop to `structure`. The value is the same.
2. Remove the `format` prop. The viewer reads the format from the text.

### 3. One effect for a selected chain and a pointed chain

Before this pull request, the viewer showed the two states in different ways:

- A pointer on a chain name made the chain pink.
- A selection of a chain made ball-and-stick atoms and an outline above the
  cartoon.

After this pull request, the viewer shows both states in one way. The colors of
the chain do not change. The other chains become 65% transparent.

- A pointer on a chain name makes the other chains transparent.
- A selection of a chain makes the other chains transparent. The effect
  continues after the pointer moves away.
- While the pointer is on a chain name, that chain is the opaque chain. Then the
  viewer shows the selection again.
- A hidden chain cannot be the opaque chain. If you hide the selected chain, all
  chains become opaque.
- `disableChainHighlightOnHover` stops the effect for the pointer only. A
  selection still makes the other chains transparent.

The camera moves to the selected chain, as before. All chains stay in the view,
because the viewer makes the other chains transparent.

A selection of residues does not change. Mol* draws ball-and-stick atoms for the
residues and for the residues near them.

The sequence panel shows a selected chain only in the caption of the chain. The
panel does not make the residues of the chain brighter.

### Tests

- The `@czi-sds/data-viz` package passes 285 tests.
- ESLint, TypeScript, and the namespace check pass.
- The documentation tests and the playground tests pass.

### How to test in Storybook

1. Run `yarn start`.
2. Open the `WithMMCIFFormat` story. Confirm that the viewer shows the structure.
3. Open the `Complex` story.
4. Put the pointer on the name of chain B in the legend. Confirm that chain A
   becomes transparent.
5. Move the pointer off the name. Confirm that chain A becomes opaque.
6. Open the `ComplexWithChainSelected` story. Confirm that chain A is transparent
   and chain B is opaque.
7. Click the name of chain B. Confirm that all chains become opaque.
8. Click one residue in the sequence panel. Confirm that Mol* draws
   ball-and-stick atoms for the residue.

## Checklist

- [x] Test Story in Storybook
- [x] Tests written
- [x] Semantic Variables from `defaultTheme.ts` used wherever possible
- [ ] If updating an existing component, depreciate flag has been used where
      necessary — no consumer uses this component, so the API changes have no
      deprecation step
- [x] Documents updated


tihuan (approved):
LGTM! Thanks so much, Masoud 💯 🙌

PR #1193, opened 2026-09-17
feat: data viz subpath exports

itsliana:
## What

Adds per-component subpath exports to `@czi-sds/data-viz`:

```ts
import { GenomeTrack } from "@czi-sds/data-viz/GenomeTrack";
```
## Why
The package entry is a barrel, and a barrel is a single module to a bundler — importing one component reaches every other component's module-level code.

  | before | after
-- | -- | --
bundle | 4960 KiB | 1665 KiB
gzipped | 1385 KiB | 449 KiB
gzip chunks | 4 | 1

<br class="Apple-interchange-newline">
The second commit marks molstar/rxjs as optional peers, dropping 105 MB of install for consumers that never render a structure.

## Notes
- The barrel is unchanged for existing consumers — and now split across the same chunks, so bundlers can drop unused ones there too.
- exports keeps a ./dist/* passthrough so any existing deep import still resolves.
- echarts/lodash are deliberately left forced: they're in both dependencies and peerDependencies, so dependencies installs them regardless. Recovering that 66 MB would break HeatmapChart consumers who never declared echarts — separate change.
- rootDir added to the namespace-check config; required once exports exists (TS2209).

## Checklist

- [ ] Test Story in Storybook
- [ ] Tests written
- [ ] Documents updated


tihuan (approved):
Thanks so much for doing this 🎉 ! Didn't know `styled(...)` calls survive tree shaking 💀 Cutting down bundle size is a lifesaver 🏆 !

tihuan (commented):
Not sure if this helps, but Gemini offers using a babel plugin to annotate the styled components to help with tree shaking 🌲 👀 

In case it helps cut down manually exporting components going forward!

Image

itsliana:
Correction to my own PR description first, because I had the mechanism wrong — it is **not** the `styled(...)` calls that survive tree shaking. Sorry for planting that 😅

What actually pins Mol\* is module-scope Mol\* API calls in the ProteinStructureViewer chunk:

```js
const CHAIN_HIGHLIGHT_COLOR = Color.fromRgb(200, 45, 149);
const CONNECTING_BOND = MolScriptBuilder.core.flags.hasAny([...]);
const DEFAULT_PLDDT_COLOR = Color.fromRgb(128, 128, 128);
const PlddtColoring = CustomElementProperty.create({ ... });
```

Five statements, none pure-annotated. They're referenced by the component code, so the import edges stay open.

On the babel plugin - two things I measured in the built chunk:

- rolldown **already** emits 86 `@__PURE__` annotations, but every one is on a JSX factory call (`jsx`/`jsxs`) inside a function body, which does nothing for module-scope shaking
- `styled(...)`: **40 calls, 0 annotated** — so the gap you're pointing at is genuinely there

But it wouldn't have replaced this PR:

1. styled components aren't what held Mol\* in, so annotating them changes nothing here
2. annotating the actual culprits only half-works — `Color.fromRgb` and the `MolScriptBuilder` expression are genuinely pure, but `CustomElementProperty.create({...})` registers a custom property. That's a real side effect you can't honestly mark `@__PURE__`, and that one statement keeps Mol\* in the graph by itself
3. there's no Babel in this pipeline — rolldown transpiles TS/JSX natively via Oxc — so adopting it means adding a Babel pass or getting Oxc to emit the annotations

Still think it's worth doing as separate bundle-hygiene work: 40 unannotated `styled()` calls in one chunk is real, and it would help anyone importing the barrel instead of a subpath. Complementary to the split rather than an alternative. Happy to file a follow-up issue if you want it tracked.

tihuan:
Ah thanks for the investigation 😄 💡 TIL! 

Just created a PR to add sideEffects config too, which seems to help as well

And I think that's probably all we need to get most of the benefits!

https://github.com/chanzuckerberg/sci-components/pull/1195

PTAL thank you 🙏

PR #1195, opened 2026-09-17
fix: declare sideEffects on components and data-viz

tihuan:
## What

Declares `sideEffects` on the two published packages that were missing it:

```jsonc
// packages/data-viz/package.json
"sideEffects": false

// packages/components/package.json
"sideEffects": ["**/*.css", "**/*.scss"]
```

## Why

Without the field, a bundler must assume every module in the package does work at import time, so it keeps import edges alive even when all of a module's bindings have been dropped. That capped what consumers could shake out regardless of the subpath exports added in #1193.

Measured with webpack 5, `mode: "production"`, data-viz bundled and everything else external:

entry | `sideEffects` | bytes | distinct molstar edges
-- | -- | -- | --
barrel, `HeatmapChart` only | omitted | 246,132 | 28
barrel, `HeatmapChart` only | `false` | **16,022** | **0**
subpath `/HeatmapChart` | `false` | 15,919 | 0
barrel, `ProteinStructureViewer` | `false` | 212,088 | 28

A barrel import of one component now lands within 103 bytes of the equivalent subpath import. The last row is the correctness control — a consumer who actually imports the viewer still gets all 28 molstar edges, so the shaking is not over-aggressive.

## Why `components` gets the array form and not `false`

`dist` ships `variables.css` and `variables.scss` as copied assets (rolldown treats the `.css`/`.scss` imports in `src/index.ts` as empty modules, and `rollup-plugin-copy` emits the real files), and consumers import those paths directly. A blanket `false` covers them too — and since a stylesheet has no exports, it gets dropped whole:

`sideEffects` | `bundle.css` | `--sds-` custom properties
-- | -- | --
omitted | 123,577 bytes | 1462
`false` | **0 — nothing emitted** | **0**
`["**/*.css","**/*.scss"]` | 123,577 bytes | 1462

Dev builds do not tree-shake, so that breakage would only have surfaced in a downstream production build.

## Notes

- Audited before declaring: no bare side-effect imports survive into either `dist`, and `css` from `@emotion/css` (Tooltip, TooltipCondensed, Tabs) is only ever called inside functions, never at module scope — so there is no style injection at import time.
- `@czi-sds/mcp` and `@czi-sds/icons` already declared `sideEffects: false` and ship no style assets, so these two were the only gaps.
- Subpath exports from #1193 are still doing work that this does not replace: CJS `require()` consumers get no tree shaking at all, and not every toolchain honors `sideEffects`.
- Worth filing upstream separately: molstar ships no `sideEffects` field. It does not block this change — webpack deletes data-viz's own viewer modules wholesale and the molstar edges live inside them — but it does mean molstar's internals cannot be shaken for consumers who do import the viewer.

## Verification

- `yarn build` — clean across all 3 projects
- `yarn test` — 154 tests pass
- `yarn lint`, `yarn namespace-check` — pass
- webpack 5 production fixtures for both tables above, including an A/B that confirms `false` on `components` drops all 1462 design tokens

## Checklist

- [ ] ~~Default Story in Storybook~~ — N/A, packaging only, no component change
- [ ] ~~Test Story in Storybook~~ — N/A
- [ ] ~~Tests written~~ — N/A; verified by bundler fixtures rather than unit tests (see Verification)
- [ ] ~~Semantic Variables from `defaultTheme.ts` used wherever possible~~ — N/A
- [ ] ~~If updating an existing component, depreciate flag has been used where necessary~~ — N/A
- [ ] Documents updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)


masoudmanson (approved):
Aah thanks for the fix @tihuan! it LGTM 🤩🥳🚀

PR #1196, opened 2026-09-18
fix(data-viz): force residue tooltip text color inline

tihuan:
## Summary

Hovering a residue in `ProteinStructureViewer`'s sequence panel shows an **empty white bubble** instead of the enlarged letter, in a consuming app (Biohub Platform) but not in Storybook.

`ResidueTooltip` pins its own surface with an **inline** style, but leaves the paired text color to the SDS `Tooltip`'s Emotion class (`color: textPrimary`, `packages/components/src/core/Tooltip/style.ts:57`). Where that class loses the cascade, MUI's own default `color: palette.common.white` wins — the color MUI pairs with its default *dark* surface — so the letter is painted white on the forced-white surface. Storybook has no CSS layers and no App Router Emotion cache, so the class wins there and the bug never shows.

Measured in the consuming app:

```
text:  "N"                  <- the label was always correct
color: rgb(255, 255, 255)
bg:    rgb(255, 255, 255)
```

A component that forces a background at inline precedence while leaving the foreground at class precedence can always be split by its host, so the pair now travels together.

Consumers cannot work around this themselves: the sequence panel renders inside Mol*'s own React root, "outside the consumer's provider tree" (`utils/theme.ts:8-11`), so neither their `ThemeProvider` nor a `MuiTooltip` `styleOverrides` reaches it. An unlayered global CSS override would reach it (the tooltip portals to `body`) but would repaint every other tooltip in the app.

## Changes

- `ResidueTooltip` sets `color` inline alongside `backgroundColor`, from `getSemanticColors({ theme })?.base?.textPrimary`.
- New `__tests__/residueTooltip.test.tsx`, covering light and dark.

The test asserts the **inline** style, not the computed one — the computed value is satisfied by the very class rule that loses the cascade, so a computed-style assertion passes while the letter is invisible. Verified by reverting the one-line fix: 2 tests fail, and they pass again with it.

## Test plan

- [x] `vitest run` for `@czi-sds/data-viz` — 25 files, 373 tests passing
- [x] `tsc --noEmit`, `eslint --quiet`, `stylelint`, `namespace-check` — all clean
- [x] New test fails without the fix (confirmed by temporarily removing it)
- [ ] Reviewer check in Storybook that the tooltip is unchanged there (it should be — this only raises the precedence of a color that already resolved to the same token)

## Notes for reviewers

Ruled out before landing on this: molstar version skew (`SequenceWrapper`'s `residueLabel` / `isHighlighted` / `isSelected` / `isFocused` are identical between 4.12.1 and 5.11.0), duplicate `@czi-sds/components` or `@emotion/react` copies, the feature missing from the published bundle, and theme resolution failure.

Unrelated observation: the root `CLAUDE.md` still documents Jest, but the packages run Vitest now.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


masoudmanson (approved):
LGTM, thanks Timmy :D

